PR #6, opened 2017-04-11
CALCITE-1539:Enable proxy access to Avatica server for third party on…

Wancy:
… behalf of end users

### DIFF
--- a/server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
+++ b/server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
@@ -48,11 +48,12 @@ public class RemoteUserExtractHttpServerTest extends HttpAuthBase {
   private static HttpServer server;
   private static String url1;
   private static String url2;
+  private static String url3;
 
   @BeforeClass public static void startServer() throws Exception {
 
     final String userPropertiesFile = BasicAuthHttpServerTest.class
-            .getResource("/auth-users.properties").getFile();
+            .getResource("/auth-users.properties").getFile().replace("%20", " ");
     assertNotNull("Could not find properties file for basic auth users", userPropertiesFile);
 
     // Create a LocalService around HSQLDB
@@ -83,16 +84,15 @@ public class RemoteUserExtractHttpServerTest extends HttpAuthBase {
                                                 Callable<T> action) throws Exception {
 
             if (remoteUserName.equals("USER4")) {
-              throw new RuntimeException("USER4 is a disallowed user");
-            } else if (remoteUserName.equals("USER2")) {
-              return action.call();
+              throw new RuntimeException("USER4 is a disallowed user!");
             } else {
-              throw new RuntimeException("Unknown user.");
+              return action.call();
             }
           }
 
           @Override public RemoteUserExtractor getRemoteUserExtractor() {
             return new RemoteUserExtractor() {
+              boolean useDoAs = false;
               HttpQueryStringParameterRemoteUserExtractor paramRemoteUserExtractor =
                       new HttpQueryStringParameterRemoteUserExtractor();
               HttpRequestRemoteUserExtractor requestRemoteUserExtractor =
@@ -100,14 +100,16 @@ public class RemoteUserExtractHttpServerTest extends HttpAuthBase {
 
               @Override public String extract(HttpServletRequest request)
                   throws RemoteUserExtractionException {
-                if (request.getParameter("doAs") != null) {
-                  String doAsUser = request.getParameter("doAs");
-                  LOG.info("doAsUser is " + doAsUser);
+                LOG.info(request.toString());
+                if (request.getParameter("useDoAs") != null
+                    && request.getParameter("useDoAs").equals("true")) {
+                  useDoAs = true;
+                }
+                if (useDoAs) {
                   return paramRemoteUserExtractor.extract(request);
                 } else {
-                  return "USER2";
+                  throw new RemoteUserExtractionException("HTTP/401");
                 }
-
               }
             };
           }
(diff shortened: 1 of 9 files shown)


server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+import static org.hamcrest.core.StringContains.containsString;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+/**
+ * Test class for HTTP Basic authentication.
+ */
+public class RemoteUserExtractHttpServerTest extends HttpAuthBase {
```
joshelser: It would be good to use the JUnit parameterized annotation to run this test against both JSON and PROTOBUF server endpoints.

server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+  private static final ConnectionSpec CONNECTION_SPEC = ConnectionSpec.HSQLDB;
+  private static HttpServer server;
+  private static String url1;
+  private static String url2;
+
+  @BeforeClass public static void startServer() throws Exception {
+
+    final String userPropertiesFile = BasicAuthHttpServerTest.class
+            .getResource("/auth-users.properties").getFile();
```
joshelser: See CALCITE-1752 -- this will fail if the path to this file contains any characters that would be url encoded.

server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
```diff
@@ -0,0 +1,175 @@
...
+              throw new RuntimeException("USER4 is a disallowed user");
+            } else if (remoteUserName.equals("USER2")) {
+              return action.call();
+            } else {
+              throw new RuntimeException("Unknown user.");
+            }
+          }
+
+          @Override public RemoteUserExtractor getRemoteUserExtractor() {
+            return new RemoteUserExtractor() {
```
joshelser: I think this test could be expanded to cover two more cases:

1. Test that when the server is configured to extract a user in a manner in which it wasn't provided by the client (e.g. server expects doAs=, but the client doesn't include that)
2. Test what happens when the remote user could not be extracted. The client should see an HTTP/401 (as opposed to an HTTP/403 or HTTP/5XX).

server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  @AfterClass public static void stopServer() throws Exception {
+    if (null != server) {
+      server.stop();
+    }
+  }
+
+  @Test public void testUserWithAllowedDoAsRole() throws Exception {
+    // Disallowed by avatica
```
joshelser: This comment should be removed.

joshelser:
@Wancy  just wanted to make sure you saw my comments on this PR.

server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
Wancy: Seems JUnit parameterized.parameter are not applicable in BeforeClass method, because BeforeClass is static while different parameters will start a new class instance, so the params will show null pointer error.

server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
Wancy: 1. I tried to mock the real usage by adding ?useDoAs=true param in the url indicating if the server expects doAs or not, is that something you are expecting?
2. One weird thing is if I use request.getRemoteUser() it will return null in this case, so I just "hardcode" to return USER2 in the extractor. Not sure if using basic authentication influences this because this extraction behavior only happens in spnego scenario, but the test cases are basic auth and I manually add avaticaServerConfiguration to force execute this code.

server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
joshelser: Yeah, parameterized and BeforeClass/AfterClass are tricky to work around, but there are examples of this working in other test cases. I can clean this up if you'd prefer.

server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
joshelser: > I tried to mock the real usage by adding ?useDoAs=true param in the url indicating if the server expects doAs or not, is that something you are expecting?

Ah, right. This might require changing some client implementation to test it properly.

I don't expect the client to provide this (after normally the proxy would tack this on), but we should still have tests that verify functionality when the server requires a doAs=user and the client does not provide it (should be some HTTP/401 or HTTP/403).

> One weird thing is if I use request.getRemoteUser() it will return null in this case

This sounds like a bug in your code then. Your implementation should throw an exception if that method returns null. These tests should be as terse/explicitly as possible.

Wancy: Yeah I realize it is missing in the implementation of throwing exception if the extracted user is null, added it in the new commit. Thanks.

server/src/test/java/org/apache/calcite/avatica/server/RemoteUserExtractHttpServerTest.java
Wancy: I tried different ways like adding the startserver method in the constructor, but get different errors. It will be great if you offer some help on this, thanks :)

risdenk:
@joshelser and @Wancy  - Chance this didn't get closed when PR #8 got merged?

joshelser:
@risdenk ya, just needs to be closed. @Wancy can do it if she has a second, otherwise, I'll close it via commit-msg on the next commit I push :)

Wancy:
Thanks @risdenk and @joshelser!

PR #8, opened 2017-04-25
CALCITE-1539 Support remote user extraction from HTTP query string parameters

joshelser:
Builds on @wancy's support in https://github.com/apache/calcite-avatica/pull/6

* Changes the end-to-end test to run JSON and PROTOBUF serializations via JUnit parameterization
* Actually sets the proper HTTP status codes for 401 and 403 (previously, just returned strings as HTTP/500)
* Created a new exception to enable the above
* Simplified some things in the test case ("useDoAs" was contrived).


joshelser:
@wancy let me know what you think of my changes in 0e77205.

I'd also prefer to get down to just one commit, so I'll likely squash my commit into your's and update your commit message, if that's OK.

Wancy:
Hi @joshelser thanks for adding these changes, I like the idea of adding RemoteUserDisallowedException, yeah you can go ahead squash your commits with mine.

PR #12, opened 2017-07-26
CALCITE-1904 Allow SSL hostname verification to be turned off

joshelser:


joshelser:
Any chance you could take a quick glance at this, @julianhyde? I'd appreciate a set of eyes just because :)

julianhyde:
+1

It's cosmetic, but I'd shorten `SslCertificateHostnameVerificationConfigurable` to say `HostnameVerificationConfigurable` to avoid very long lines.

joshelser:
Thanks, Julian!

> It's cosmetic, but I'd shorten SslCertificateHostnameVerificationConfigurable to say HostnameVerificationConfigurable to avoid very long lines.

Yeah, I think I agree with ya. I had tried to be terse early when starting the change, but let it lapse by the end (the enum being `HostnameVerification` and not `SslCertificateHostnameVerification`).

PR #14, opened 2017-08-03
CALCITE-1915 Work around a Jetty where the SPNEGO challenge is not …

joshelser:
…sent


joshelser:
Need to write a unit test.

PR #15, opened 2017-08-03
CALCITE-1922 Allow kerberos v5 OID in SPNEGO authentication

joshelser:


joshelser:
As much as I'd like to write a unit test for this, I think it's a little beyond my current knowledge. I don't presently have the time to spend writing an HTTP client capable of SPNEGO authentication by hand.

Most of this change is copy-paste from jetty-9.2.15 (our current dependency).

PR #17, opened 2017-10-06
CALCITE-2003 Remove global synchronization on openConnection

joshelser:


joshelser:
@julianhyde might you know why Travis is killing the maven build after just a couple of minutes? Maybe I'm missing some more configuration?

https://travis-ci.org/apache/calcite-avatica/builds/284317333

julianhyde:
no idea. looks a bit like this: https://github.com/travis-ci/travis-ci/issues/8172

risdenk:
@joshelser Did you see https://github.com/apache/calcite/pull/542? Any chance it is a surefire issue?

joshelser:
@risdenk no, I did miss that one! Let me try that quick. Otherwise, I'll just leave the privileged=true tweak on since that "worked"

joshelser:
Ok, that is unrelated. We only have the failsafe-plugin configured to do parallel tests in Avatica, not surefire-plugin too. Thanks for the suggestion though, Kevin!

As I understand it, doing the `sudo: required` just increases the amount of time it takes for tests to start running (the things it actually provides aren't needed for Avatica). Given the relative low-velocity of commits to Avatica, I think that's a fine tradeoff for stable test reports.

PR #23, opened 2018-01-27
[CALCITE-508] wrapping RuntimeException into SQLException in Cursor.Accessor getAccessor method, added unit tests

asolimando:
The PR in summary:

1) I have basically followed the suggestion of wrapping RuntimeExceptions into SQLException into accessors by handling this into the getObject methods, as consistently invoked under the hood by the accessors themselves.

2) I have added some unit test for checking for error while accessing columns for a cursor not initialized (i.e., next() was never called), and for checking that the same operations were ok after calling next() on the cursor.

3) Minor refactoring of tests in "AvaticaResultSetThrowsSqlExceptionTest.java" by the addition of an auxiliary method.

I still have one doubt:
Line 71 in IteratorCursor.java: throw new NoSuchElementException("Invalid cursor position");
Would it make more sense to directly throw this exception as an instance of SQLException (as it will end up anyway, being a RuntimeException) or not?

### DIFF
--- a/core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
+++ b/core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
@@ -53,6 +53,21 @@ public class AvaticaResultSetThrowsSqlExceptionTest {
     }
   }
 
+  /**
+   * Auxiliary method returning a result set on a test table.
+   * @return a result set on a test table.
+   * @throws SQLException in case of database error
+   */
+  private ResultSet getResultSet() throws SQLException {
+    Properties properties = new Properties();
+    properties.setProperty("timeZone", "GMT");
+
+    final TestDriver driver = new TestDriver();
+    final Connection connection = driver.connect("jdbc:test", properties);
+
+    return connection.createStatement().executeQuery("SELECT * FROM TABLE");
+  }
+
   @Test
   public void testPrevious() throws SQLException {
     Properties properties = new Properties();
@@ -60,8 +75,8 @@ public class AvaticaResultSetThrowsSqlExceptionTest {
 
     final TestDriver driver = new TestDriver();
     try (Connection connection = driver.connect("jdbc:test", properties);
-         ResultSet resultSet =
-             connection.createStatement().executeQuery("SELECT * FROM TABLE")) {
+       ResultSet resultSet =
+               connection.createStatement().executeQuery("SELECT * FROM TABLE")) {
       thrown.expect(SQLFeatureNotSupportedException.class);
       resultSet.previous();
     }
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
```diff
@@ -75,11 +90,118 @@ public void testUpdateNull() throws SQLException {
...
+  @Test
+  public void testCommonCursorStates() throws SQLException {
+    final ResultSet resultSet = getResultSet();
+
+    // right after statement execution, result set is before first row
+    assert resultSet.isBeforeFirst();
+
+    // retrieve each row until the last one
+    while (!resultSet.isAfterLast()) {
+      assert resultSet.next() != resultSet.isAfterLast();
```
vlsi: Please avoid using Java asserts in the test code. Test failures will be hard to understand/analyze in case of assert failures.

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
```diff
@@ -75,11 +90,118 @@ public void testUpdateNull() throws SQLException {
...
 
+  @Test
+  public void testCommonCursorStates() throws SQLException {
+    final ResultSet resultSet = getResultSet();
+
+    // right after statement execution, result set is before first row
+    assert resultSet.isBeforeFirst();
+
+    // retrieve each row until the last one
+    while (!resultSet.isAfterLast()) {
```
vlsi: Infinite loop is possible. Please add a counter

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
```diff
@@ -75,11 +90,118 @@ public void testUpdateNull() throws SQLException {
...
+      // we have not called next, so each column getter should throw SQLException
+      for (int i = 1; i <= resultSet.getMetaData().getColumnCount(); i++) {
+        //System.out.println(resultSet.getMetaData().getColumnTypeName(i));
+        assert getColumn(resultSet, i, true);
+      }
+    }
+  }
+
+  @Test
+  public void testGetColumnsAfterNext() throws SQLException {
```
vlsi: Is there `getColumnsAfterLast` test?

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
```diff
@@ -75,11 +90,118 @@ public void testUpdateNull() throws SQLException {
...
+        resultSet.getTime(index);      // TIME
+        break;
+      case 11:
+        resultSet.getTimestamp(index); // TIMESTAMP
+        break;
+      default:
+        resultSet.getObject(index);
+      }
+    } catch (SQLException e) {
+      if (!shouldThrow) {
```
vlsi: What bothers me is the test verifies a mere presence of `SQLException`. Is it possible to test for the exact exception the test expects? E.g. `getSQLState`, `getErrorCode`, 

`

getErrorCode
--

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
asolimando: I have added a limit of at most 2 rows, that is sufficient for the test at hand. 
Do you see any better way to enforce a reasonable limit?
There is no way to know beforehand the size of the result set, you can just check the return value of next and isAfterLast (this last was my previous attempt).

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
asolimando: I have replaced them with the JUnit counterparts (e.g., Assert.assertTrue), this would do?

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
asolimando: I totally agree with you, but if I am not wrong the official documentation for JDBC SQL states and error codes can only be purchased from ANSI and XOpen, and this is not an option for me. 

Open documentation is available from different vendors, but they often map error codes differently from the standard and among them. 

I would be happy to raise (and consequently test) specific error codes if you can suggest a source where I can freely access them.

asolimando:
Hi @vlsi,
I have update the PR according to your suggestions (modulo the more precise error codes, as discussed above), can you please have a look?

I haven't managed to rebase my two commits for CALCITE-508 due to the intermixed commit for CALCITE-2083, is that ok? If not, how do you usually handle such situations?

asolimando:
Hello @vlsi, @julianhyde, can you have a look at the PR? It would be nice to make it before the release of Avatica 1.16.0, so CALCITE-508 could be carried out also on Calcite side.

Thanks!

julianhyde:
Sorry for the delay. I've made https://issues.apache.org/jira/browse/CALCITE-508 a dependency of the release https://issues.apache.org/jira/browse/CALCITE-2182, so we'll get to it in time.

PR #26, opened 2018-03-06
[CALCITE-1520] Implement isValid for AvaticaConnection

risdenk:
* Changed helper to a static variable on the connection so Mockito could instantiate it. Since the underlying helper INSTANCE is static, I didn't think this would be an issue. Most of the line changes come from fixing Checkstyle for static final variable being uppercase.
* Implemented isValid without fully checking the underlying connection is actually valid. Does check if the connection is closed.
* Added test for isValid for negative timeout and isClosed checks.

There didn't seem to be a way to check if there is a valid connection to the Avatica server. I thought about adding to the service or trying to get to the underlying HttpClient. However, none of these expose a timeout that would be reasonable. This implementation although not complete is a start in the right direction.


joshelser:
Change looks good to me, @risdenk.

My only ask would be if you could add a test case to `RemoveDriverTest` in the server module? Something simple/high-level that doesn't rely on mocking (e.g. create a connection, use it, check `isValid`, close the connection, ensure `isValid() == false`.

risdenk:
Sure will add that today hopefully

risdenk:
@joshelser - thanks for the quick review. Added a test case to `RemoteDriverTest`

risdenk:
Looks like the test failure is related to JDK7 and bumping the hsqldb version?

joshelser:
Yeah, not your problem. Will try to merge this sucka in today. Thanks for the PR!

PR #29, opened 2018-03-11
[CALCITE-2210] Remove oraclejdk7 and add oraclejdk9 to .travis.yml

risdenk:
The only supported Travis CI Oracle verions are oraclejdk8 and oraclejdk9. For more details see:
https://docs.travis-ci.com/user/reference/trusty/#JVM-(Clojure%2C-Groovy%2C-Java%2C-Scala)-images

IBM JDK is not installed by default unless you install it. Removing the commented out IBM JDK as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,11 +20,11 @@ language: java
 matrix:
   fast_finish: true
   include:
-    - jdk: oraclejdk8
-    - jdk: openjdk7
-      # Travis appears to have broken IBM JDK. Can only get a JRE.
-      # Needs to be investigate.
-      #    - jdk: ibmjava8
+    - env: IMAGE=maven:3-jdk-10
+    # Workaround: https://github.com/docker-library/openjdk/issues/145
+    - env: IMAGE=maven:3-jdk-9-slim
+    - env: IMAGE=maven:3-jdk-8
+    - env: IMAGE=maven:3-ibmjava-8
 branches:
   only:
     - master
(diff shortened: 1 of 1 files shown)


joshelser:
> IBM JDK is not installed by default unless you install it. Removing the commented out IBM JDK as well.

I'd like us to get back to running against IBM JDK too. Do you have insight into how to get Travis to do this auto-magically? It was working at one point and then... stopped :)

Fix is good by me.

risdenk:
From what I read it looks like you can install any JDK and then use jdk_switcher to switch to it. I haven't looked much closer than that yet.

risdenk:
Seems like there is info about IBM JDK here: https://github.com/travis-ci/travis-ci/issues/2682

https://github.com/travis-ci/packer-templates/pull/527 talks about how we might be able to enable this with a more recent image specified in the .travis.yml. I can try to play around and see.

risdenk:
@joshelser - How about the solution of using a Docker image on Travis? The Docker image [`maven:ibmjava`](https://hub.docker.com/_/maven/) is built on IBM Java. I figured why not build all in Docker containers since Maven Docker has jdk 8, 9, and 10.

joshelser:
I'm actually ok with that. I imagine there isn't much difference in speed. Let me look at this..

joshelser:
I like it. It passes, and doesn't seem to add an exorbitant amount of time to the build.

PR #32, opened 2018-03-18
[CALCITE-2218] Fix AvaticaResultSet#getRow()

laurentgo:
According to JDBC spec/Javadoc, ResultSet#getRow() method is 1-indexed
based, but AvaticaResultSet#getRow() is 0-indexed based.


core/src/main/java/org/apache/calcite/avatica/AvaticaResultSet.java
```diff
@@ -398,15 +403,15 @@ public BigDecimal getBigDecimal(String columnLabel) throws SQLException {
   }
 
   public boolean isBeforeFirst() throws SQLException {
-    return row < 0;
```
risdenk: Couldn't this just be `row < 1` to avoid introducing the `beforeFirst` variable?

risdenk: Ah makes sense. Thanks for the clarification.

risdenk:
Changes look good and has a test. Looking into merging this soon.

laurentgo:
Thanks for checking. I can take care of the merging too.

PR #33, opened 2018-03-19
[CALCITE-2219] Check if resource is closed, and throws

laurentgo:
According to JDBC spec, most methods from Connection, Statement,
PreparedStatement and ResultSet should throw an exception if the
resource is closed.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/AvaticaConnection.java
+++ b/core/src/main/java/org/apache/calcite/avatica/AvaticaConnection.java
@@ -154,9 +154,15 @@ public abstract class AvaticaConnection implements Connection {
     this.meta.openConnection(handle, OpenConnectionRequest.serializeProperties(info));
   }
 
+  protected void checkOpen() throws SQLException {
+    if (isClosed()) {
+      throw HELPER.closed();
+    }
+  }
   // Connection methods
 
   public AvaticaStatement createStatement() throws SQLException {
+    checkOpen();
     //noinspection MagicConstant
     return createStatement(ResultSet.TYPE_FORWARD_ONLY,
         ResultSet.CONCUR_READ_ONLY,
@@ -164,6 +170,7 @@ public abstract class AvaticaConnection implements Connection {
   }
 
   public PreparedStatement prepareStatement(String sql) throws SQLException {
+    checkOpen();
     //noinspection MagicConstant
     return prepareStatement(
         sql, ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY,
@@ -179,18 +186,22 @@ public abstract class AvaticaConnection implements Connection {
   }
 
   public void setAutoCommit(boolean autoCommit) throws SQLException {
+    checkOpen();
     meta.connectionSync(handle, new ConnectionPropertiesImpl().setAutoCommit(autoCommit));
   }
 
   public boolean getAutoCommit() throws SQLException {
+    checkOpen();
     return unbox(sync().isAutoCommit(), true);
   }
 
   public void commit() throws SQLException {
+    checkOpen();
     meta.commit(handle);
   }
 
   public void rollback() throws SQLException {
+    checkOpen();
     meta.rollback(handle);
   }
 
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/apache/calcite/avatica/AvaticaConnection.java
```diff
@@ -276,14 +298,17 @@ public CallableStatement prepareCall(
   }
 
   public Map<String, Class<?>> getTypeMap() throws SQLException {
-    throw HELPER.unsupported();
+    checkOpen();
+    // Always return an empty map since custom types is not supported
+    return new HashMap<>();
```
risdenk: Would it be more correct to return `SQLFeatureNotSupportedException`? I don't think it matters either way but below is from the javadoc for Connection.

> SQLFeatureNotSupportedException - if the JDBC driver does not support this method

core/src/main/java/org/apache/calcite/avatica/AvaticaConnection.java
```diff
@@ -392,14 +420,17 @@ public void setClientInfo(Properties properties)
   }
 
   public String getClientInfo(String name) throws SQLException {
-    throw HELPER.unsupported();
+    checkOpen();
+    return null;
```
risdenk: This could be `return this.getClientInfo().getProperty(name)` now that `getClientInfo()` returns a `Properties` object. This would remove the need for the added `checkOpen()` and explicit `return null`.

core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
```diff
@@ -268,18 +278,20 @@ protected void close_() {
...
 
   public int getMaxFieldSize() throws SQLException {
-    throw connection.HELPER.unsupported();
+    checkOpen();
+    return 0;
   }
 
   public void setMaxFieldSize(int max) throws SQLException {
-    throw connection.HELPER.unsupported();
+    throw AvaticaConnection.HELPER.unsupported();
```
risdenk: Would it make sense to make this be a noop instead of returning unsupported now that `getMaxFieldSize()` is set to always return 0.

laurentgo: not if the value is set to something different from 0, if not supported

core/src/main/java/org/apache/calcite/avatica/AvaticaConnection.java
laurentgo: yes, probably more correct. I think I got the truncated version of the javadoc on this one. I will revert.

core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
risdenk: Ok that makes sense. Thanks!

joshelser:
@laurentgo any chance I could request that you write a test case for one of this scenarios? Maybe the one in which lead you to find that this bug existed?
Otherwise, looks good to me too.

laurentgo:
Not really triggered by a bug, but more by the extra code in Drill and Dremio JDBC drivers to implements the JDBC spec. Let me see if can reuse the same strategy as the Drill code (helper class to iterate over methods and check that the methods throws the correct exception).

There are also lots of methods where unsupported is thrown instead of a sqlexception for closed. I wonder if the correct thing to do would be to check first for closed, and then throw unsupported, and if it's fine for now.

risdenk:
@laurentgo - Any plans to add a test or few for this?

laurentgo:
yes, still my plan, just need to find some time to wrap it up.

laurentgo:
@risdenk @joshelser any comment regarding the unit tests?

risdenk:
Whoops definitely missed there was a commit added to this. Not sure it sends an email :( I can take a look.

core/src/test/java/org/apache/calcite/avatica/AvaticaClosedCheckHelper.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.stream.Stream;
+
+import static org.hamcrest.CoreMatchers.endsWith;
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+/**
+ * Helper class for tests checkign close behavior
```
risdenk: sp nit: checking

core/src/test/java/org/apache/calcite/avatica/AvaticaClosedConnectionTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Tests for {@code AvaticaConnection} relative to close behavior
+ */
+@RunWith(Parameterized.class)
+public class AvaticaClosedConnectionTest {
+
+  private static final Predicate<? super Method> METHOD_FILTER = new Predicate<Method>() {
+    public boolean test(Method m) {
+      // Skip methods not overriden by Avatica
```
risdenk: sp nit: overridden

core/src/test/java/org/apache/calcite/avatica/AvaticaClosedPreparedStatementTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+/**
+ * Tests for {@code AvaticaPreparedStatement} relative to close behavior
+ */
+@RunWith(Parameterized.class)
+public class AvaticaClosedPreparedStatementTest {
+  private static final Predicate<? super Method> METHOD_FILTER = new Predicate<Method>() {
+    public boolean test(Method m) {
+      // Skip methods not overriden by Avatica
```
risdenk: sp nit: overridden

core/src/test/java/org/apache/calcite/avatica/AvaticaClosedResultSetTest.java
```diff
@@ -0,0 +1,141 @@
...
+import static org.mockito.Mockito.mock;
+
+/**
+ * Tests for {@code AvaticaResultSet} relative to close behavior
+ */
+@RunWith(Parameterized.class)
+public class AvaticaClosedResultSetTest {
+  private static final Predicate<? super Method> METHOD_FILTER = new Predicate<Method>() {
+    public boolean test(Method m) {
+      // Skip methods not overriden by Avatica
```
risdenk: sp nit: overridden

laurentgo:
If I merge the checkstyle patch first, I might be able to revert to my lambda implementation :)

F21:
Friendly ping @laurentgo

PR #37, opened 2018-04-20
[CALCITE-1914] DOAP file for Calcite Avatica

risdenk:


risdenk:
@joshelser - FYI put you as maintainer. Not sure that is 100% correct but you seem to be involved from the beginning of Avatica. Based on http://calcite.apache.org/news/2017/03/31/new-avatica-repository/

PR #40, opened 2018-04-20
[CALCITE-2269] Enable errorprone checking

risdenk:


risdenk:
Looks like this is going to fail on JDK 11 :( See https://github.com/google/error-prone/issues/860

risdenk:
The error-prone fix for JDK 11 is in there now. Looking at rebasing this on master.

risdenk:
Apparently JDK 11 support has a few more issues that should be resolved in errorprone 2.3.2.

risdenk:
This has issues on JDK 12 now :/ Seems to be a moving target.

F21:
@risdenk Do you think this can make it into the next release?

risdenk:
I think so. I'll change up how it is enabled to make sure everything is in a profile. I don't think we should have it enabled by default for building Avatica. This was a question on another project.

risdenk:
Added profile `ErrorProne` that will enable the Error Prone checking on demand. I did not add the JDK 8 stuff since we will eventually move away from JDK 8. Added to `.travis.yml` to check error prone on JDK 11 only.

risdenk:
Travis is having a few issues today. Will wait to merge after Travis finishes to make sure the Error Prone profile is triggered correctly for JDK 11.

PR #41, opened 2018-04-21
[CALCITE-2255] Add JDK 11 to Travis CI

risdenk:
Moved from jdk 9 slim since there is a better way to handle the SSL workaround with truststore password.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,9 @@ language: java
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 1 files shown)


risdenk:
Woot tests passing for JDK 8, 9, 10, and 11!

risdenk:
Rebased on top of master and this will now check javadocs. I'm expected this to fail due to [LANG-1384](https://issues.apache.org/jira/browse/LANG-1384) which makes javadocs fail on JDK 11.

PR #42, opened 2018-04-21
[CALCITE-2272] Apache Parent POM 19, JDK 10 Surefire, and JDK 10 javadoc

risdenk:


risdenk:
Rebased on master to confirm everything still good.

PR #46, opened 2018-05-01
[CALCITE-2284] Allow Jetty Server to be customized before startup

aaraujo:
This allows downstream Avatica users to have fine grained control over
SSL configuration without exposing uncommon settings through the
HttpServer Builder.

### DIFF
--- a/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
+++ b/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
@@ -48,7 +48,10 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.security.Principal;
 import java.security.PrivilegedAction;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
 import java.util.Set;
@@ -75,6 +78,7 @@ public class HttpServer {
   private final AvaticaServerConfiguration config;
   private final Subject subject;
   private final SslContextFactory sslFactory;
+  private final List<ServerCustomizer<Server>> serverCustomizers;
   private final int maxAllowedHeaderSize;
 
   @Deprecated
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/calcite/avatica/server/ServerCustomizer.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Callback for customizing a Jetty {@link Server}.
+ */
+public interface ServerCustomizer {
+  /**
+   * Customize the server during initialization.
+   * @param server The server to customize
+   */
+  void customize(Server server);
```
joshelser: I think this should be changed -- perhaps parameterize this class with the concrete `Server` type. Avatica's API should not strictly depend on an implementation details (in this case, that Jetty is being used). We don't want users to know that Jetty is the webserver used behind the scenes.

e.g.

```java
public interface ServerCustomizer<T> {
  void customize(T server);
}
```

Making this change would also require some changes to go into the HttpServer.Builder class to make sure that users know exactly what they're getting. I think deprecating `HttpServer` and making it `JettyHttpServer` instead is the right avenue (or something along those lines).

Does that make sense?

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
```diff
@@ -75,6 +76,7 @@
   private final AvaticaServerConfiguration config;
   private final Subject subject;
   private final SslContextFactory sslFactory;
+  private final List<ServerCustomizer> serverCustomizers;
```
joshelser: I think we should force this to be non-null (as you did in the Builder). An empty list as the "default" value passed in through the constructors would let us eliminate the if-null check elsewhere.

server/src/test/java/org/apache/calcite/avatica/server/HttpServerCustomizerTest.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @AfterClass public static void stopServer() {
+    if (null != server) {
+      server.stop();
+    }
+  }
+
+  @Test public void serverCustomizersInvoked() {
+    verify(mockCustomizer2, times(1)).customize(any(Server.class));
```
joshelser: nit: `times(1)` is unnecessary, I think.

Nice test addition though!!

aaraujo:
WDYT of the API change I pushed @joshelser?.

aaraujo:
I upgraded phoenix-queryserver locally to a patched Avatica that contained this API change. Module tests passed. Can you think of additional compat tests I should run?

aaraujo:
> I'm not sure if Java can handle this with binary compatibility or not.

Java seems to provide backwards compat for the raw API by adding additional methods to class files during compile time. This is consistent with my local testing.

> If Java doesn't insulate us from this change, we'll have to make this change in a new HttpServer+Builder pair

Happy to add a new pair, to be safe, if that eases your concern.

PR #48, opened 2018-05-19
CALCITE-2294 Allow customization for AvaticaServerConfiguration for p…

karanmehta93:
…lugging new authentication mechanisms

@joshelser @julianhyde @aaraujo 
Please review!

@apurtell FYI

### DIFF
--- a/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
+++ b/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
@@ -239,18 +239,8 @@ public class HttpServer {
     server.setConnectors(new Connector[] { connector });
 
     // Default to using the handler that was passed in
-    final HandlerList handlerList = new HandlerList();
-    Handler avaticaHandler = handler;
+    configureHandlers(securityHandler);
 
-    // Wrap the provided handler for security if we made one
-    if (null != securityHandler) {
-      securityHandler.setHandler(handler);
-      avaticaHandler = securityHandler;
-    }
-
-    handlerList.setHandlers(new Handler[] {avaticaHandler, new DefaultHandler()});
-
-    server.setHandler(handlerList);
     // Apply server customizers
     for (ServerCustomizer<Server> customizer : this.serverCustomizers) {
       customizer.customize(server);
@@ -274,7 +264,22 @@ public class HttpServer {
     }
   }
 
-  private ServerConnector getConnector() {
+  protected void configureHandlers(ConstraintSecurityHandler securityHandler) {
+    final HandlerList handlerList = new HandlerList();
+    Handler avaticaHandler = handler;
+
+    // Wrap the provided handler for security if we made one
+    if (null != securityHandler) {
+      securityHandler.setHandler(handler);
+      avaticaHandler = securityHandler;
+    }
+
+    handlerList.setHandlers(new Handler[] {avaticaHandler, new DefaultHandler()});
+
+    server.setHandler(handlerList);
+  }
+
+  protected ServerConnector getConnector() {
     HttpConnectionFactory factory = new HttpConnectionFactory();
     factory.getHttpConfiguration().setRequestHeaderSize(maxAllowedHeaderSize);
 
@@ -467,6 +472,8 @@ public class HttpServer {
 
     // The maximum size in bytes of an http header the server will read (64KB)
     private int maxAllowedHeaderSize = MAX_ALLOWED_HEADER_SIZE;
+    private AvaticaServerConfiguration serverConfig;
+    private Subject subject;
 
     public Builder() {}
 
@@ -659,6 +666,12 @@ public class HttpServer {
       return withAuthentication(AuthenticationType.DIGEST, properties, allowedRoles);
     }
 
+    public Builder<T> withCustomAuthentication(AvaticaServerConfiguration config) {
+      this.authenticationType = AuthenticationType.CUSTOM;
+      this.serverConfig = config;
+      return this;
+    }
+
     private Builder<T> withAuthentication(AuthenticationType authType, String properties,
         String[] allowedRoles) {
       this.loginServiceRealm = "Avatica";
@@ -721,8 +734,6 @@ public class HttpServer {
      */
     @SuppressWarnings("unchecked")
     public HttpServer build() {
-      final AvaticaServerConfiguration serverConfig;
-      final Subject subject;
       switch (authenticationType) {
       case NONE:
         serverConfig = null;
@@ -747,6 +758,10 @@ public class HttpServer {
         }
         serverConfig = buildSpnegoConfiguration(this);
         break;
+      case CUSTOM:
+        serverConfig = buildCustomConfiguration(this);
+        subject = null;
+        break;
       default:
         throw new IllegalArgumentException("Unhandled AuthenticationType");
       }
@@ -769,6 +784,10 @@ public class HttpServer {
 
       return new HttpServer(port, handler, serverConfig, subject, sslFactory, jettyCustomizers,
           maxAllowedHeaderSize);
+    }
+
+    private AvaticaServerConfiguration buildCustomConfiguration(Builder<T> tBuilder) {
+      return tBuilder.serverConfig;
     }
 
     /**
--- a/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
+++ b/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
@@ -688,6 +688,16 @@ public class HttpServer {
       return withAuthentication(AuthenticationType.DIGEST, properties, allowedRoles);
     }
 
+    /**
+     * Configures the server to use CUSTOM authentication mechanism, which can allow users to
+     * combine benefits of multiple auth methods. See <code>CustomAuthHttpServerTest</code> for
+     * examples on how to use it.
+     * Note: Default ServerConnectors and Handlers will NOT be used.
+     * Customize them directly using instances <code>{@link ServerCustomizer}</code>
+     * @param config AvaticaServerConfiguration implementation that configures various details
+     *      about the authentication mechanism for <code>{@link HttpServer}</code>
+     * @return <code>this</code>
+     */
     public Builder<T> withCustomAuthentication(AvaticaServerConfiguration config) {
       this.authenticationType = AuthenticationType.CUSTOM;
       this.serverConfig = config;
--- a/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
+++ b/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
@@ -220,7 +220,7 @@ public class HttpServer {
     if (null != this.config && AuthenticationType.CUSTOM == config.getAuthenticationType()) {
       if (null != handler || null != sslFactory) {
         throw new IllegalStateException("Handlers and SSLFactory cannot be configured with "
-                + "HTTPServer Builder when using CUSTOM Authentication Type.");
+                + "the HTTPServer Builder when using CUSTOM Authentication Type.");
       }
     } else {
       serverConnector = configureServerConnector();
@@ -251,7 +251,9 @@ public class HttpServer {
         throw new RuntimeException(e);
       }
     } else if (0 == server.getConnectors().length) {
-      LOG.warn("No server connectors have been configured for this Avatica server.");
+      String error = "No server connectors have been configured for this Avatica server";
+      LOG.error(error);
+      throw new RuntimeException(error);
     }
   }
 
--- a/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
+++ b/server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
@@ -215,34 +215,21 @@ public class HttpServer {
     server = new Server(threadPool);
     server.manage(threadPool);
 
-    final ServerConnector connector = configureConnector(getConnector(), port);
-    ConstraintSecurityHandler securityHandler = null;
-
-    if (null != this.config) {
-      switch (config.getAuthenticationType()) {
-      case SPNEGO:
-        // Get the Handler for SPNEGO authentication
-        securityHandler = configureSpnego(server, connector, this.config);
-        break;
-      case BASIC:
-        securityHandler = configureBasicAuthentication(server, connector, config);
-        break;
-      case DIGEST:
-        securityHandler = configureDigestAuthentication(server, connector, config);
-        break;
-      default:
-        // Pass
-        break;
+    ServerConnector serverConnector = null;
+    HandlerList handlerList = null;
+    if (null != this.config && AuthenticationType.CUSTOM == config.getAuthenticationType()) {
+      if (null != handler || null != sslFactory) {
+        throw new IllegalStateException("Handlers and SSLFactory cannot be configured with "
+                + "HTTPServer Builder when using CUSTOM Authentication Type.");
       }
+    } else {
+      serverConnector = configureServerConnector();
+      handlerList = configureHandlers();
     }
-
-    server.setConnectors(new Connector[] { connector });
-
-    // Default to using the handler that was passed in
-    configureHandlers(securityHandler);
 
     // Apply server customizers
     for (ServerCustomizer<Server> customizer : this.serverCustomizers) {
+      LOG.info("Customizing server with customizer: " + customizer.getClass());
       customizer.customize(server);
     }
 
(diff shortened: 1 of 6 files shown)


karanmehta93:
Added a new `CUSTOM` authentication type for users to plugin `AvaticaServerConfiguration`.
Tests in the class `CustomAuthHttpServerTest.java` tests such scenarios.

I am fixing the check style issues as well, meanwhile please have a look.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
```diff
@@ -747,6 +758,10 @@ public HttpServer build() {
         }
         serverConfig = buildSpnegoConfiguration(this);
         break;
+      case CUSTOM:
+        serverConfig = buildCustomConfiguration(this);
```
joshelser: This indirection seems unnecessary. Let's not add this unless there's something this method is supporting :)

server/src/test/java/org/apache/calcite/avatica/server/CustomAuthHttpServerTest.java
```diff
@@ -0,0 +1,300 @@
...
+    final JdbcMeta jdbcMeta = new JdbcMeta(CONNECTION_SPEC.url,
+            CONNECTION_SPEC.username, CONNECTION_SPEC.password);
+    LocalService service = new LocalService(jdbcMeta);
+
+    server = new HttpServer.Builder()
+            .withCustomAuthentication(new CustomImpersonationConfig())
+            .withHandler(service, Driver.Serialization.PROTOBUF)
+            .withPort(0)
+            .build();
+    server.start();
```
joshelser: You need to ensure any server you start is stopped after the test method completes

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
```diff
@@ -239,18 +239,8 @@ protected void internalStart() {
...
     // Default to using the handler that was passed in
-    final HandlerList handlerList = new HandlerList();
-    Handler avaticaHandler = handler;
-
-    // Wrap the provided handler for security if we made one
-    if (null != securityHandler) {
-      securityHandler.setHandler(handler);
-      avaticaHandler = securityHandler;
-    }
+    configureHandlers(securityHandler);
```
joshelser: At this point, `securityHandler` is null because you don't have a case statement for `CUSTOM` a few lines above.

This means that your CUSTOM AuthenticationType is obtusely "unimplemented". There isn't any clarity in here for me to know what security methods on the `AvaticaServerConfiguration` will actually have effect, and which will not.

For example, I may try to configure the login realm via the `ASC`, but I don't think that would actually be interrogated. When constructing the Jetty server.

karanmehta93: Agreed, we can add a clause that `CUSTOM` config has to be used with appropriate `ServerCustomizer` to create the server. I can document that appropriately. We can also restrict users to use `ServerCustomizer`  with `CUSTOM` type for the following reason.

`ServerCustomizer` was added to provide additional options for configuration to the server. Since we pass in the `Jetty server` object directly, one can directly remove all the connectors , security handlers and other relevant properties from the server, which can potentially lead to unexpected bugs.

Here (at SFDC), we need a way to customize the servers so that we can configure our own `SSLContextFactory` and `ServerConnector` on the handler (based off some internal code for MTLS auth). Having customizers allow us to do that, however we don't have a mechanism to bypass all default `AvaticaServerConfiguration` logic with the methods provided. We also need to support impersonation, which is only allowed with `buildSpnegoConfiguration()` method, however we don't want to use SPNEGO. This `CUSTOM` method of authentication opens up a lot of doors to customize both authZ and authN as per requirements. As shown in example, we can also use this up for adding impersonation with BasicAuth, which is currently not supported. We still want to use the same handlers as the core logic remains same.

Let me know your thoughts on this one!

I agree that this is not the _best_ way of doing this. I believe that `AvaticaServerConfiguration` interface should provide methods to authenticate the user. For example, the method `isUserPermitted()` is called first inside the `AvaticaProtobufHandler` as well as `AvaticaJsonHandler` classes, where the authentication happens only for the case of SPNEGO. Ideally this method should be wrapped around every handler that is implemented, since I can potentially add a new Handler and forget to call this method at the start. Also, if we add new auth methods, we need to add if-else statements over here to ensure that we cover all the cases. I feel that in such a case `AvaticaServerConfiguration` should be self-sufficient to provide an interface method that can handle the part based on its type. We can use that wrapper to also handle common pieces of code in these handlers like exception handling or impersonation allowance based on serverConfig.

If you can point me to the code that helps in such wrapping, I am happy to do the refactoring.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
karanmehta93: This method helps in getting the `AvaticaServerConfiguration` object from the builder. Why do you not want to have it?

server/src/test/java/org/apache/calcite/avatica/server/CustomAuthHttpServerTest.java
karanmehta93: The `@After` annotated method on Line 292, ensure that server is stopped after every test case.
I could not use the `@Before` method to start the server, since every test required a different type of server. I have tried to refactor the code so that we can reuse the methods as required. Let me know if it requires further refactoring.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
```diff
@@ -274,7 +264,22 @@ protected void internalStart() {
     }
   }
 
-  private ServerConnector getConnector() {
+  protected void configureHandlers(ConstraintSecurityHandler securityHandler) {
```
apurtell: Is this type too narrow? Would there be other types of handlers that someone customizing the server would want to add to the list?

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
apurtell: See my other comment. While what we care about is customizing security handlers, since we have to think about generic interface. Is it enough just to say CUSTOM means that the user will supply a list of handlers including whatever security handler they want to use, so we don't build anything canned ourselves? We are pretty close to this already.

karanmehta93: I agree, in order to comply with `CUSTOM` as the authentication method, we need to also allow users to specify that. The builder allows users to add their handlers, but still wraps them around with the `securityHandler` which gets configured based on the auth type. 

A better of approaching this would be to bypass all these code for `CUSTOM` security config, so that it doesn't bring any expectations to the user. What do you think?

karanmehta93:
`AvaticaServerConfiguration` has a few methods in place such as `getAllowedRoles` or `getHashLoginServiceRealm`, which are only applicable to certain types of auth methods. I feel that is breaks the interface level assumptions. We can rather have a class here that keeps these properties directly (rather than assigning them via Builder and passing them along) and various auth methods can inherit from it. This can help make `AvaticaServerConfiguration` self-sufficient for any security checks that we want to have in place. 

Please advise on better ways of implementing this in terms of good code practices.

FYI @vincentpoon

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
apurtell: > A better of approaching this would be to bypass all these code for CUSTOM security config, so that it doesn't bring any expectations to the user.

Yes, exactly

karanmehta93: I can go ahead and add that change here, @joshelser Please advice.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
joshelser: Are you sure? I'm not sure this is doing anything. If I'm reading this right, this line is effectively:

```java
serverConfig = this.serverConfig;
```

You already did this in `withCustomAuthentication`, so I'm not sure what the intent is to do here. Please tell me if I'm missing something (I'm admittedly not looking at this in an editor, so it might just be me)

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
joshelser: I like your suggestion, @apurtell! I was a bit reticent to suggest that as, like you imply, it's a bit more work to set up all of those handlers on your own. However, this would also fit nicely into the work that @aaraujo has already done (the parameterized Jetty-specific stuff for the `Builder` class).

I think that providing "well-known authentication scenarios" is a nice thing that we can do (e.g. basic, digest, spengo), and then leave `CUSTOM` for the "let me drive" scenario where you set it all up yourself.

I'm still a bit curious if there's another "nicer" abstraction we could provide from an API perspective that isn't full CUSTOM. For example, are there more abstractions which would help us put the boilerplate into the HttpServer.Builder and let you inject your SFDC specific authn/authz decisions? I am just a little ignorant on the specifics of how your SFDC-specific thing works. You can also just tell me to drop it and I'll stop poking :)

joshelser: > one can directly remove all the connectors , security handlers and other relevant properties from the server, which can potentially lead to unexpected bugs.

Yeah, this is the angle I'm coming from. I can appreciate the flexibility of `CUSTOM`, but am worried if that's "safe" for the common-folk.

> I believe that AvaticaServerConfiguration interface should provide methods to authenticate the user.

I think that makes sense. My biggest worry about `AvaticaServerConfiguration` is that it's started to become a "dumping ground". Every time I've added a new authn option, it results in an implementation specific option that is only valid "sometimes". Not saying that what you're doing here requires an immediate re-write, just trying to give you my perspective :)

> Ideally this method should be wrapped around every handler that is implemented, since I can potentially add a new Handler and forget to call this method at the start

+1

> If you can point me to the code that helps in such wrapping, I am happy to do the refactoring.

`o.a.c.a.server.HandlerFactory` is what instantiates the `Avatica*Handler` classes now. I think you'd just need to pass the `ASConfiguration` (or some interface which just has the authz/impersonation related methods -- the Avatica handlers shouldn't need to know all HTTP server configuration details) into the construction of the Handler.

This would be some great follow-on.

server/src/test/java/org/apache/calcite/avatica/server/CustomAuthHttpServerTest.java
karanmehta93: +1, Will make it consistent.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
karanmehta93: > I like your suggestion, @apurtell! I was a bit reticent to suggest that as, like you imply, it's a bit more work to set up all of those handlers on your own

Agreed and that's probably the reason why I forgot abstracting them out as well :)

> I'm still a bit curious if there's another "nicer" abstraction we could provide from an API perspective that isn't full CUSTOM. For example, are there more abstractions which would help us put the boilerplate into the HttpServer.Builder and let you inject your SFDC specific authn/authz decisions? I am just a little ignorant on the specifics of how your SFDC-specific thing works. You can also just tell me to drop it and I'll stop poking :)

@joshelser I initially wanted to do that, however `AvaticaServerConfiguration` is not abstracted out completely. I will file a Jira to refactor the code and can make it better/generic. MTLS as a type of authentication is a slight change to TLS part (where we configure `ServerConnectors` to accept client certs). Even I want to push as much code upstream as possible :)

karanmehta93:
> I think if you're happy with turning this into CUSTOM auth type into a "I'll configure the server myself", that would be great

As of now, yes. I will update the PR with your comments.

> There's some other good stuff mentioned here, but I don't want us to get bogged down talking about things we want to do and confuse what you're actually trying to accomplish now.

I will file Jira items to refactor the code.

> Thanks for your work here already. It's great!

Thank you! Appreciate you for taking out time for review and comments.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
karanmehta93: Nice catch @joshelser , That's correct, its effectively the same line when I had a closer look on it.
I think I just created that method to keep it in sync with other Auth methods above. 
What do you want me to do? Remove the case from the switch?

joshelser: I would leave the switch case, but replace what you have presently with a comment telling future-us that we don't need to do any configuration building for `CUSTOM`.

karanmehta93: Okay, Just to clarify, I will keep the same code in place and add the comments.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
karanmehta93: Sure will file a follow up Jira for that.

karanmehta93:
@joshelser Do you want me to put the edits to the security page as a part of this Jira itself?
All I have to do is to modify the `security.md` file, is that correct?

karanmehta93:
@joshelser The doc is still remaining, however your comments have been addressed. Let me know if you require further changes.
@apurtell The handler config is also customized now.
@aaraujo Please review.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
```diff
@@ -215,34 +215,21 @@ protected void internalStart() {
...
-        break;
-      default:
-        // Pass
-        break;
+    ServerConnector serverConnector = null;
+    HandlerList handlerList = null;
+    if (null != this.config && AuthenticationType.CUSTOM == config.getAuthenticationType()) {
+      if (null != handler || null != sslFactory) {
+        throw new IllegalStateException("Handlers and SSLFactory cannot be configured with "
+                + "HTTPServer Builder when using CUSTOM Authentication Type.");
```
joshelser: nit: "with this Builder" or "with the HttpServer Builder"

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
```diff
@@ -215,34 +215,21 @@ protected void internalStart() {
...
     }
 
-    server.setConnectors(new Connector[] { connector });
-
-    // Default to using the handler that was passed in
-    configureHandlers(securityHandler);
-
     // Apply server customizers
     for (ServerCustomizer<Server> customizer : this.serverCustomizers) {
+      LOG.info("Customizing server with customizer: " + customizer.getClass());
```
joshelser: Nice addition.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
```diff
@@ -251,35 +238,70 @@ protected void internalStart() {
...
-      throw new RuntimeException(e);
+      // Set the information about the address for this server
+      try {
+        this.handler.setServerRpcMetadata(createRpcServerMetadata(serverConnector));
+      } catch (UnknownHostException e) {
+        // Failed to do the DNS lookup, bail out.
+        throw new RuntimeException(e);
+      }
+    } else if (0 == server.getConnectors().length) {
+      LOG.warn("No server connectors have been configured for this Avatica server.");
```
joshelser: This sounds like it should be thrown as a RuntimeException to me. If there are no connectors, this HttpServer couldn't possibly be doing what it's intended to do.

server/src/test/java/org/apache/calcite/avatica/server/CustomAuthHttpServerTest.java
```diff
@@ -175,63 +126,165 @@ public void testCustomBasicImpersonationConfigWithDisallowedUser() throws Except
...
+            .withHandler(Mockito.mock(AvaticaHandler.class))
+            .withPort(0)
+            .build();
+    try {
+      server.start();
+      fail("Expected an exception");
+    } catch (IllegalStateException e) {
+      String assertString = "Handlers and SSLFactory cannot be configured with "
+              + "HTTPServer Builder when using CUSTOM Authentication Type";
+      assertThat(e.getMessage(), containsString(assertString));
```
joshelser: I'm not a fan of tests that assert a certain error message. IMO, if it's important enough to validate that we hit this exact code path (and not another similar), we should have a named exception.

I think just validating that the `ISE` was thrown is sufficient.

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
```diff
@@ -659,6 +686,12 @@ public Builder withRemoteUserExtractor(RemoteUserExtractor remoteUserExtractor)
       return withAuthentication(AuthenticationType.DIGEST, properties, allowedRoles);
     }
 
+    public Builder<T> withCustomAuthentication(AvaticaServerConfiguration config) {
```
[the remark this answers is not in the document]
joshelser: I think Karan was just following what I had elsewhere in here (`withDigestAuthentication(String, String[])`, `withBasicAuthentication(String, String[])`, and the inconsistently named `withSpnego(String, String, String[])`).

Sounds like a great follow-on to me :)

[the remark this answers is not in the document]
karanmehta93: Unsure here about the comment. What do you think looks neater?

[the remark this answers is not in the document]
joshelser: IMO, what you have now is fine. I think what Andrew suggested would be a nice improvement, but we would have to make sure we don't break compatibility among other things. That would just be good to separate out and handle later.

joshelser:
> Do you want me to put the edits to the security page as a part of this Jira itself? All I have to do is to modify the security.md file, is that correct?

If you can, that'd be great and that is the right place to put it :)

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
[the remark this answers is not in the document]
karanmehta93: Great. I feel that `withServerConfiguration()` method name is not a very good representative of what it does internally and hence would like to prefer my current approach only. I can probably overload the private method `withAuthentication()` so that we should not cause any trouble, but I think its good for now.

joshelser:
Cool, this looks good to me after you latest patchset, @karanmehta93. Will try to get this merged today.

PR #49, opened 2018-05-23
[CALCITE-2322] Add fetch size support to connection url and JDBC state…

kminder:
Adds the support for both a URL connection param fetch_size and the JDBD Statement API setFetchSize.  This can be used to tune performance for queries that return a large number of rows.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
+++ b/core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
@@ -36,6 +36,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
  */
 public abstract class AvaticaStatement
     implements Statement {
+
   /** The default value for {@link Statement#getFetchSize()}. */
   public static final int DEFAULT_FETCH_SIZE = 100;
 
@@ -108,6 +109,7 @@ public abstract class AvaticaStatement
     this.resultSetType = resultSetType;
     this.resultSetConcurrency = resultSetConcurrency;
     this.resultSetHoldability = resultSetHoldability;
+    this.fetchSize = connection.config().fetchSize(); // Default to connection config fetch size.
     this.signature = signature;
     this.closed = false;
     if (h == null) {
--- a/core/src/main/java/org/apache/calcite/avatica/BuiltInConnectionProperty.java
+++ b/core/src/main/java/org/apache/calcite/avatica/BuiltInConnectionProperty.java
@@ -77,7 +77,10 @@ public enum BuiltInConnectionProperty implements ConnectionProperty {
   TRUSTSTORE_PASSWORD("truststore_password", Type.STRING, null, false),
 
   HOSTNAME_VERIFICATION("hostname_verification", Type.ENUM, HostnameVerification.STRICT,
-      HostnameVerification.class, false);
+      HostnameVerification.class, false),
+
+  /** Fetch size limit, default is 100 rows. */
+  FETCH_SIZE("fetch_size", Type.NUMBER, AvaticaStatement.DEFAULT_FETCH_SIZE, false);
 
   private final String camelName;
   private final Type type;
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
```diff
@@ -108,6 +109,7 @@ protected AvaticaStatement(AvaticaConnection connection,
     this.resultSetType = resultSetType;
     this.resultSetConcurrency = resultSetConcurrency;
     this.resultSetHoldability = resultSetHoldability;
+    this.fetchSize = connection.config().fetchSize(); // Default to connection config fetch size.
```
laurentgo: let's remove the default value set for the field declaration

core/src/main/java/org/apache/calcite/avatica/BuiltInConnectionProperty.java
```diff
@@ -77,7 +77,10 @@
   TRUSTSTORE_PASSWORD("truststore_password", Type.STRING, null, false),
 
   HOSTNAME_VERIFICATION("hostname_verification", Type.ENUM, HostnameVerification.STRICT,
-      HostnameVerification.class, false);
+      HostnameVerification.class, false),
+
+  /** Fetch size limit, default is 100 rows. */
+  FETCH_SIZE("fetch_size", Type.NUMBER, AvaticaStatement.DEFAULT_FETCH_SIZE, false);
```
laurentgo: I would recommend to name it `DEFAULT_FETCH_SIZE`

kminder: I'm not clear on your suggestion.  Are you suggesting that the url connection parameter be named "DEFAULT_FETCH_SIZE" or as you suggest above "defaultFetchSize".  You mention having looked at a few other drivers I've looked too and don't see much consistency.  I chose "fetch_size" because that seemed consistent with the other parameters.  Would "default_fetch_size" be agreeable?

core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
kminder: I'm not clear on what you are suggesting here.  Are you suggesting this be defaulted to 0 and have the real value resolved in getFetchSize()?  I went with the approach I did because a setting made via the connection url is accessed through the connection object.  The connection reference isn't stored within AvaticaStatement so the connection value needs to be propagated down somehow.  You are correct that if the value is explicitly set to 0 either via url param or via Statement.setFetchSize or ResultSet.setFetchSize I have no idea what will happen.  All I can say is that I didn't change that behavior with this change.  It will do whatever is would have done before.

core/src/main/java/org/apache/calcite/avatica/BuiltInConnectionProperty.java
laurentgo: yes, default_fetch_size for the jdbc property, which means the enum become DEFAULT_FETCH_SIZE (that was my point :) )

F21:
@kminder Do you think it would be possible to get this change in by the end of this week?

F21:
@kminder We are planning to release Avatica 1.13.0 soon. Would you be able to take another look at this PR?

zacMode:
Hi, is there a reason this was never merged? 

We're currently struggling with this and would love to see this make it into the next release. Thanks!

F21:
@zacMode Would you be interested in taking over this PR? If so, please go ahead and open a new PR based on this and I can close out this one. There's further discussion on JIRA as well: https://issues.apache.org/jira/browse/CALCITE-2322

zacMode:
> @zacMode Would you be interested in taking over this PR? If so, please go ahead and open a new PR based on this and I can close out this one. There's further discussion on JIRA as well: https://issues.apache.org/jira/browse/CALCITE-2322

I'm open to it. I've read over this PR and the corresponding JIRA ticket, but It's not clear what's left to do?

F21:
I think there's currently some discussion over the use of camelCase vs snake_case. As the last comment in JIRA is almost 3 years old, I think it would be best to ping the JIRA ticket to confirm whether camelCase or snake_case should be used and go from there.

zacMode:
Got it, will do. Thanks!

zabetak:
Closing this PR as there is a follow up on https://github.com/apache/calcite-avatica/pull/148.

PR #51, opened 2018-05-29
[CALCITE-2333] Stop releasing zips

F21:
This updates the website and maven configs to stop releasing zips for future releases.

The site was tested locally and works fine.

Can someone verify that I have updated the pom.xml files correctly?


F21:
I've moved the first commit to master (it's really trivial), but it's still showing in the PR for some reason.

Pinging @joshelser for a second look.

michaelmior:
Yeah, I'm not really concerned about the other commit. Seems like a bug in GitHub that it hasn't gone away.

F21:
Thanks, Josh. I'll give that a shot!

PR #53, opened 2018-05-30
Bump HSQLDB to 2.4.1

F21:
This is a minor release, so there should not be any breaking changes.

Changelog shows mostly bug fixes and a few minor enhancements: http://hsqldb.org/doc/2.0/changelist_2_0.txt


F21:
The tests pass, but ccing @joshelser and @risdenk for a review.

joshelser:
If tests are passing, no concerns from me, @F21 !

PR #55, opened 2018-06-05
[CALCITE-2341] Fix javadoc plugin incompatibility

laurentgo:
Fix the javadoc plugin incompatibility with JDK10 or higher but
updating the plugin version from 3.0.0 to 3.0.1


risdenk:
:+1: Thanks!

PR #56, opened 2018-06-06
[CALCITE-2352] Update to checkstyle 6.18

laurentgo:
Update project to maven-checkstyle-plugin 3.0.0/checkstyle 6.18.
Fix all new reported violations.


risdenk:
Changes look reasonable at a glance. Are there any compatibility issues with changing the method signatures? I can run a check on this if that helps.

laurentgo:
From a public API perspective, there should not be: checkstyle only complains if constructor has a greater visibility than the class itself. The only exception has been the factory which is accessed through reflection (and the code assumes the constructor is public)

laurentgo:
Thanks for the check! unless any objection, I'll merge it tomorrow (cc @F21 )!

laurentgo:
Merged in acb675de97b9b0743c09368820a770e2ceda05f8

PR #57, opened 2018-06-08
CALCITE-2285 Support client cert keystore for Avatica Client

karanmehta93:
@joshelser @aaraujo 
Please review.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
+++ b/core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
@@ -81,7 +81,8 @@ public class AvaticaCommonsHttpClientImpl implements AvaticaHttpClient,
   protected final URI uri;
   protected BasicAuthCache authCache;
   protected CloseableHttpClient client;
-  PoolingHttpClientConnectionManager pool;
+  protected Registry<ConnectionSocketFactory> socketFactoryRegistry;
+  protected PoolingHttpClientConnectionManager pool;
 
   protected UsernamePasswordCredentials credentials = null;
   protected CredentialsProvider credentialsProvider = null;
@@ -102,8 +103,8 @@ public class AvaticaCommonsHttpClientImpl implements AvaticaHttpClient,
   }
 
   private void initializeClient() {
-    Registry<ConnectionSocketFactory> configureSocketFactory = configureSocketFactories();
-    configureConnectionPool(configureSocketFactory);
+    socketFactoryRegistry = this.configureSocketFactories();
+    configureConnectionPool(socketFactoryRegistry);
     this.authCache = new BasicAuthCache();
     // A single thread-safe HttpClient, pooling connections via the ConnectionManager
     this.client = HttpClients.custom().setConnectionManager(pool).build();
@@ -136,18 +137,7 @@ public class AvaticaCommonsHttpClientImpl implements AvaticaHttpClient,
     }
 
     try {
-      SSLContextBuilder sslContextBuilder = SSLContexts.custom();
-
-      if (null != truststore && null != truststorePassword) {
-        sslContextBuilder.loadTrustMaterial(truststore, truststorePassword.toCharArray());
-      }
-
-      if (null != keystore && null != keystorePassword && null != keyPassword) {
-        sslContextBuilder.loadKeyMaterial(keystore,
-                keystorePassword.toCharArray(), keyPassword.toCharArray());
-      }
-
-      SSLContext sslContext = sslContextBuilder.build();
+      SSLContext sslContext = getSSLContext();
       final HostnameVerifier verifier = getHostnameVerifier(hostnameVerification);
       SSLConnectionSocketFactory sslFactory = new SSLConnectionSocketFactory(sslContext, verifier);
       registryBuilder.register("https", sslFactory);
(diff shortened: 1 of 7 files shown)


aaraujo:
Lgtm, but can you add tests for the new config paths @karanmehta93?

karanmehta93:
@aaraujo 
I did manual testing here. I will update the PR with tests. There are no tests covering the truststore code path, so I will have to figure out a way of adding tests there.

F21:
Thanks for contributing, @karanmehta93! I'd love to get this into avatica-1.12 if possible.

karanmehta93:
@F21 @aaraujo Added tests for the new code. Please review.

F21:
ping @joshelser

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
```diff
@@ -127,11 +121,54 @@ private void initializeClient() {
...
+    RegistryBuilder<ConnectionSocketFactory> registryBuilder = RegistryBuilder.create();
+    configureHttpRegistry(registryBuilder);
+    configureHttpsRegistry(registryBuilder);
+    return registryBuilder.build();
+  }
 
-    // A single thread-safe HttpClient, pooling connections via the ConnectionManager
-    this.client = HttpClients.custom().setConnectionManager(pool).build();
+  protected void configureHttpsRegistry(RegistryBuilder<ConnectionSocketFactory> registryBuilder) {
+    if (!configureHttpsSocket) {
```
joshelser: What about looking at the scheme of the URL instead of using "one of" the TLS configuration options?

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
```diff
@@ -95,29 +103,15 @@ public AvaticaCommonsHttpClientImpl(URL url) {
...
 
-    RegistryBuilder<ConnectionSocketFactory> registryBuilder = RegistryBuilder.create();
-    registryBuilder.register("http", PlainConnectionSocketFactory.getSocketFactory());
-    // Only register the SSL factory when provided
-    if (null != sslFactory) {
-      registryBuilder.register("https", sslFactory);
-    }
-    pool = new PoolingHttpClientConnectionManager(registryBuilder.build());
+  protected void configureConnectionPool(Registry<ConnectionSocketFactory> configureSocketFactory) {
+    pool = new PoolingHttpClientConnectionManager(configureSocketFactory);
```
joshelser: meant to be `connectionSocketFactory`?

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
joshelser: Actually, if we look at the URL, then may push this check up to `configureSocketFactories()` and call only one of `configureHttpRegsitry` and `configureHttpsRegistry`?

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
karanmehta93: Oh yes, updated it locally but forgot to push it. Will do.

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
karanmehta93: I agree, but sometimes various services can have varied requirements. The best and simple case is to have either http or https, however I didn't want to change the original implementation, so I added both of them here as well.
What do you think?

joshelser: I think there's a strong coupling of HttpClient to a specific avatica server instance ( going off of memory). If that's the case, then there's no problem in making the determination as to what kind of http client we need to set up.

karanmehta93: I looked through the code. That's correct. The `HttpHost` required by `CloseableHttpClient` is created in the constructor itself. `HttpHost` is the one that uses the URL provided during the initialization. So there's a tight coupling here and there will never be a case where both of them will be together. I will update the code to configure either of these factories depending on URL prefix as `http` vs `https`. What do you think?

karanmehta93: Actually I will use `url.getProtocol()` method to handle it for me. Seems better option.

karanmehta93:
Updated the PR. All the tests pass locally.

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
karanmehta93: As FYI, I ended up using `host.getSchemeName()` where host is a `HttpPost` object. The value is populated by `url.getProtocol()` only.

core/src/test/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImplSocketFactoryTest.java
```diff
@@ -74,25 +60,51 @@
...
+    configureClient();
+  }
+
+  private void configureClient() throws Exception {
+    client = spy(new AvaticaCommonsHttpClientImpl(url));
+    // storeFile can be used as either Keystore/Truststore
+    storeFile = mock(File.class);
+    when(storeFile.exists()).thenReturn(true);
+    when(storeFile.isFile()).thenReturn(true);
+    password = new String("");
```
joshelser: Just `""` :)

joshelser:
One minor change in the test, otherwise looks good.
Seems like we have some JDK11 issues when building -- don't need to fix that right now :)

joshelser:
I think this is ready. Will try to swing by and merge if @F21 doesn't beat me to it.

F21:
@joshelser I will let you merge as the JDK 11 tests are failing on travis.

risdenk:
Looks like the JDK 11 failure is
```
[[1;31mERROR[m] error reading /root/.m2/repository/org/bouncycastle/bcprov-jdk15on/1.55/bcprov-jdk15on-1.55.jar; invalid manifest format
```

This was brought up on the Calcite dev mailing list and could be a JDK bug with JDK 11 build 17. Looks like build 18 might fix this.

Copying below for reference:

> Agree about this might be a JDK bug.
> https://bugs.openjdk.java.net/browse/JDK-8200530
> and followed up by 
> https://bugs.openjdk.java.net/browse/JDK-8204494
> I checked the Netty Manifest file and it uses /r/n.
> This should be fixed in build 18 and was broken in build 17.

joshelser:
Thanks, Kevin! I missed that

risdenk:
Tested this change with upgrading bouncycastle to 1.59 (CALCITE-2361) locally on JDK 11 build 17 and all works good. +1 to merge from me.

karanmehta93:
@joshelser I just realized, I forgot to clean up the code for `configureHttpsSocket` variable after separating the registry builder stuff in `configureSocketFactories()` for http/https. Should I put up an addendum?

joshelser:
@karanmehta93 feel free to throw up a new PR. I'll just reopen the Jira issue for now :)

karanmehta93:
https://github.com/apache/calcite-avatica/pull/61
Done. Small change only. All tests pass locally.

PR #62, opened 2018-06-27
[CALCITE-2386] Naively wire up struct support

kennknowles:
It seems like Avatica already has struct support, but not in the particular code path used by sqlline. I've tried this out for utterly elementary bits, but no more.

```
> select row(1, 2);
+--------+
| EXPR$0 |
+--------+
| {1, 2} |
+--------+
```

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/ArrayImpl.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/ArrayImpl.java
@@ -26,6 +26,7 @@ import java.sql.Types;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 /** Implementation of JDBC {@link Array}. */
 public class ArrayImpl implements Array {
@@ -224,7 +225,7 @@ public class ArrayImpl implements Array {
     ResultSet leftResultSet = left.getResultSet();
     ResultSet rightResultSet = right.getResultSet();
     while (leftResultSet.next() && rightResultSet.next()) {
-      if (!leftResultSet.getObject(1).equals(rightResultSet.getObject(1))) {
+      if (!Objects.equals(leftResultSet.getObject(1), (rightResultSet.getObject(1)))) {
         return false;
       }
     }
--- a/core/src/main/java/org/apache/calcite/avatica/util/ArrayImpl.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/ArrayImpl.java
@@ -220,6 +220,17 @@ public class ArrayImpl implements Array {
     // nothing to do
   }
 
+  public static boolean equalContents(Array left, Array right) throws SQLException {
+    ResultSet leftResultSet = left.getResultSet();
+    ResultSet rightResultSet = right.getResultSet();
+    while (leftResultSet.next() && rightResultSet.next()) {
+      if (!leftResultSet.getObject(1).equals(rightResultSet.getObject(1))) {
+        return false;
+      }
+    }
+    return !leftResultSet.next() && !rightResultSet.next();
+  }
+
   /** Factory that can create a ResultSet or Array based on a stream of values. */
   public interface Factory {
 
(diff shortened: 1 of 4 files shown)


kennknowles:
CC: @apilloud @akedin

core/src/main/java/org/apache/calcite/avatica/util/Cursor.java
```diff
@@ -122,6 +123,8 @@
 
     Array getArray() throws SQLException;
 
+    Struct getStruct() throws SQLException;
```
kennknowles: In fact, this may be extraneous.

amaliujia:
LGTM

apilloud:
LGTM

F21:
cc @joshelser

kennknowles:
Thanks @F21!

core/src/main/java/org/apache/calcite/avatica/util/ArrayImpl.java
```diff
@@ -220,6 +220,17 @@ Object getArrayData(Object o, AbstractCursor.ArrayAccessor componentAccessor)
     // nothing to do
   }
 
+  public static boolean equalContents(Array left, Array right) throws SQLException {
+    ResultSet leftResultSet = left.getResultSet();
+    ResultSet rightResultSet = right.getResultSet();
+    while (leftResultSet.next() && rightResultSet.next()) {
+      if (!leftResultSet.getObject(1).equals(rightResultSet.getObject(1))) {
```
joshelser: Please switch this over to be `Objects.equals(leftResultSet.getObject(1), rightResultSet.getObject(1))`. Just in case either of these are ever null :)

kennknowles: Done

joshelser: > Just in case either of these are ever null

And what I really meant to say was "in case `leftResultSet.getObject(1)` is ever null" 🙃

kennknowles:
(pushing fixup commits so that github comments anchored to commits don't disappear; I'll squash after)

joshelser:
Music to my ears, @kennknowles!

Take a glance at https://github.com/apache/calcite-avatica/blob/master/server/src/test/java/org/apache/calcite/avatica/remote/RemoteMetaTest.java, or maybe https://github.com/apache/calcite-avatica/blob/master/server/src/test/java/org/apache/calcite/avatica/remote/ArrayTypeTest.java. Essentially, we have higher-level, end-to-end tests in the avatica-server module, but not any clear separation about where the tests should go in there :)

Let me know if that's not enough of a pointer to get you going (PS: I still haven't checked to make sure HSQLDB supports structs enough to do what you want -- in case you run into issues).

kennknowles:
Hmm, from a read of the manual and the code, I can't really tell if HSQLDB supports structs. Lots of basic enum definitions, seemingly know explicit code to support it, and not mentioned really in the manual.

Would you be OK merging this as-is and pursuing the high-level test later? Since it is higher risk; maybe I spend an hour on it and turns out not to be supported.

joshelser:
> Would you be OK merging this as-is and pursuing the high-level test later? Since it is higher risk; maybe I spend an hour on it and turns out not to be supported.

Yup, no problem at all :) Thanks for the quick fixup you made this morning. Let me pull it down and get it in!

kennknowles:
Awesome, thanks!

kennknowles:
Just a follow-up comment here that from what I can tell HSQLDB does not support STRUCT in its DDL, but the DDL and other support in HSQLDB is essentially undocumented. The bundled `SqlTool` also does not support displaying STRUCT data, claiming it cannot be done (why not? Sqlline does?). I also checked H2 which actually has docs, and it explicitly does not list STRUCT as a supported type. This functionality is critical for Beam SQL so I will keep poking around.

joshelser:
Thanks for digging into that more, Kenn! I appreciate you going the extra mile here.

Happy we could get this into Avatica to help you out in Beam :)

PR #63, opened 2018-07-02
[CALCITE-2394] Fix TIMESTAMP accessors: test with non-default, non-zero timezone

kennknowles:
This removes some millis-since-epoch arithmetic from the accessor for a timestamp. In fact, no arithmetic should be necessary for absolute instant references on the data plane, but only during pretty printing. The issue manifests as timestamp columns always being off for any non-zero timezone locale.


julianhyde:
Please see discussion in https://issues.apache.org/jira/browse/CALCITE-2394.

kennknowles:
Got it. Closing this for bit.

PR #64, opened 2018-07-05
[CALCITE-2390] Remove uses of X509CertificateObject, deprecated in current bouncycastle

risdenk:


risdenk:
Hmmm I can reproduce the failure locally with IBM JDK 8. Not sure how the changes here could cause this since it is a separate test. Looking into it.

risdenk:
Failure has nothing to do with this change. It looks like it might be a recent change in ibmjava. Will take a look if this keeps causing problems for other PRs.

PR #66, opened 2018-08-15
[CALCITE-2467] Upgrade owasp-dependency-check maven plugin to 3.3.1

julianhyde:
Upgrade protobuf-java to 3.5.1, jackson to 2.9.6, jetty to 9.4.11.v20180605.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,10 @@ limitations under the License.
     <avatica.release.version>${project.version}</avatica.release.version>
     <version.major>1</version.major>
     <version.minor>12</version.minor>
+
+    <!-- Don't fail the build for vulnerabilities below this threshold. -->
+    <failBuildOnCVSS>8</failBuildOnCVSS>
+
     <!-- This list is in alphabetical order. -->
     <bouncycastle.version>1.59</bouncycastle.version>
     <build-helper-maven-plugin.version>3.0.0</build-helper-maven-plugin.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -776,5 +786,28 @@ limitations under the License.
...
+      <!-- Extra checks that are disabled in the regular build, enabled for
+      releases and on demand. -->
+      <id>pedantic</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.owasp</groupId>
+            <artifactId>dependency-check-maven</artifactId>
+            <configuration>
+              <failBuildOnCVSS>${failBuildOnCVSS}</failBuildOnCVSS>
```
joshelser: Do you need to propagate this into the plugin's configuration? Would certainly be nice if the plugin would just pick it up from the definition in the properties up-above (I don't know off the top of my head if it does this though)

server/src/main/java/org/apache/calcite/avatica/server/PropertyBasedSpnegoLoginService.java
```diff
@@ -60,7 +61,8 @@ public PropertyBasedSpnegoLoginService(String realm, String serverPrincipal) {
     targetNameField.set(this, serverPrincipal);
   }
 
-  @Override public UserIdentity login(String username, Object credentials) {
+  @Override public UserIdentity login(String username, Object credentials,
+      ServletRequest request) {
```
joshelser: From the looks of https://github.com/eclipse/jetty.project/commit/c23f21c761ff56f2ad4841a98083b237494442de, the method signature changed, but there were no changes of consequence in SpnegoLoginService. We're good.

pom.xml
julianhyde: Please review https://github.com/julianhyde/calcite-avatica/commit/aabf8b935cb1ceecce4337df782f4463ecefb1bd. I added a suppressions file.

PR #69, opened 2018-09-19
[CALCITE-1183] Bump Kerby dependency version and reenable AvaticaSpnegoTest

risdenk:


risdenk:
Ping @joshelser since you filed CALCITE-1183.

PR #71, opened 2018-09-20
[CALCITE-2486] Upgrade Apache parent POM to version 21

risdenk:
[CALCITE-2486] Upgrade Apache parent POM to version 21

Also upgrade
bouncycastle to 1.60,
commons-lang3 to 3.8,
docker-maven-plugin to 1.1.1,
groovy-maven-plugin to 2.1,
h2 to 1.4.197,
httpclient to 4.5.6,
httpcore to 4.4.10,
javax-servlet-api to 4.0.1,
jcommander to 1.72,
maven-assembly-plugin to 3.1.0,
maven-enforcer-plugin to 3.0.0-M2,
maven-scm-provider-plugin to 1.11.1,
mockito to 2.22.0,
os-maven-plugin to 1.6.0,
owasp-dependency-check to 3.3.2,
protobuf to 3.6.1,
protobug-maven-plugin to 0.5.1,
slf4j to 1.7.25

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@ limitations under the License.
   <parent>
     <groupId>org.apache</groupId>
     <artifactId>apache</artifactId>
-    <version>19</version>
+    <version>21</version>
   </parent>
   <groupId>org.apache.calcite.avatica</groupId>
   <artifactId>avatica-parent</artifactId>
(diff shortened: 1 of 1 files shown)


risdenk:
I'm pretty confident in most of the changes. I checked:
* `mvn -Papache-release verify`
* `mvn -Pdocker verify`
* `mvn -Ppedantic verify`

The few dependencies I want to get a double check on:
* dropwiard-metrics3 3.1.2 to 4.0.3
* javax-servlet-api 3.0.1 to 4.0.1
* protobuf 3.5.1 to 3.6.1

@joshelser Any concerns with the above dependency bumps?

risdenk:
JDK 10 failed due to Travis CI having some issues with network.

```
[[1;31mERROR[m] Plugin org.apache.maven.plugins:maven-assembly-plugin:3.1.0 or one of its dependencies could not be resolved: Could not transfer artifact org.apache.maven.plugins:maven-assembly-plugin:jar:3.1.0 from/to central (https://repo.maven.apache.org/maven2): GET request of: org/apache/maven/plugins/maven-assembly-plugin/3.1.0/maven-assembly-plugin-3.1.0.jar from central failed: Connection reset -> [1m[Help 1][m
```

joshelser:
> Any concerns with the above dependency bumps?

If you can glance at the release-notes for httpcore, httpclient, protobuf-java, and dropwizard-metrics, that would be good. Just make sure they arent' calling out anything which we need to account for downstream.

Aside: did they really release a dropwizard-metrics3 version 4.0.3? lol

risdenk:
So the `dropwizard-metrics3` thing threw me for a loop too. It looks like the maven artifact ID changed somewhere along the way (didn't track it down.) `dropwizard-metrics3` is really just a Avatica nomenclature now. Sadly the module name is `dropwizard-metrics3` instead of `dropwizard-metrics`. I can rename the version property that has `metrics3` to `metrics` but not sure what to do about the Avatica `dropwizard-metrics3` module name since that would break backwards compatibility.

For "httpcore, httpclient, protobuf-java, and dropwizard-metrics", I checked and didn't see any problems with backwards compatibility.

joshelser:
> It looks like the maven artifact ID changed somewhere along the way (didn't track it down.)

Ok. Does that mean dropwizard-metrics 4 is compatible with 3? I think I put "3" in the first name because 2 was incompatible? Or, maybe I was just concerned about it not being forward compatible. If upstream did change how they name things, should be find to change the naming in ours and just call it out at release time. I think it would only bite people who build their own -- most people are just getting this artifact transitively.

joshelser:
> For "httpcore, httpclient, protobuf-java, and dropwizard-metrics", I checked and didn't see any problems with backwards compatibility.

+1 thanks for checking this!

risdenk:
I'm going to "revert" the Dropwizard version change and handle it separately. This way it can be called out specifically in a JIRA.

risdenk:
Created CALCITE-2583 for dropwizard metrics upgrade specifically

risdenk:
Thanks @joshelser and @julianhyde

PR #72, opened 2018-09-20
[CALCITE-2583] Upgrade dropwizard metrics to 4.0.3

risdenk:
See discussion in #71  about why dropwizard metrics should get upgraded separately.

This renames the metrics3 and dropwizard3 to remove the 3 nomenclature.


risdenk:
We can alternatively just upgrade the dropwizard-metrics dependency version and live with the 4.0.3 dependency is part of `dropwizard-metrics3`. It wouldn't be the end of the world. Just slightly misleading naming.

joshelser:
> any concerns with these changes to module/package names?

Nope, what you have is fine for me. I'd just ask that you add some kind of documentation saying that we expect it to work with DWmetrics 3 and 4. README file or even the description tag in the pom would be fine. Just a breadcrumb in case someone does come looking. (feel free to just push too -- I trust that whatever you come up with will be great)

risdenk:
Thanks @joshelser. I added a line to the module description and put a more detailed description in the commit message.

PR #73, opened 2018-09-23
[CALCITE-2587] Regenerate protobuf files for 3.6.1

risdenk:


risdenk:
@joshelser would appreciate a review on this :)

risdenk:
I went back and checked the protobuf release notes for Java from 3.3.0 -> 3.6.1 to make sure there were no surprises. I had previously only checked 3.5.1 -> 3.6.1 for the upgrade. There doesn't seem to be any major changes that would affect backwards compatibility.

The biggest note is as of 3.5.0 proto3 is preserving unknown fields which should be fine. There are steps to wrap the parser to remove the unknown fields if necessary (detailed in the release notes).

risdenk:
I also spot checked the java files that were changed and most of the changes are adding `@Override` and the handling for unknown fields. Nothing crazy in there changed.

PR #74, opened 2018-09-25
[CALCITE-2594] Ensure forbiddenapis and maven-compiler use the correct JDK version

risdenk:


pom.xml
```diff
@@ -410,8 +415,8 @@ limitations under the License.
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>8</source>
-          <target>8</target>
+          <source>${maven.compiler.source}</source>
```
joshelser: I think these are superfluous when you have `maven.compiler.source` and `maven.compiler.target` set as you do.

risdenk: Yea I think so too. I just wanted to be sure we set the same version in one spot. Afraid someone comes along later and updates the 8 to a 9/10/11 and doesn't change the maven.compiler.source/maven.compiler.target properties.

joshelser: Maybe drop them here, and leave a comment not to re-add?

risdenk: Yup that works. I'll make that change.

risdenk:
Thanks for the review @joshelser!

PR #82, opened 2019-01-08
[CALCITE-2776] Wrong value when accessing struct types with one attribute (Stamatis Zampetakis)

zabetak:
1. Remove special treatment for struct types with one attribute in AbstractCursor#createAccessor.

2. Add parameterized unit test for testing struct accessors using multiple internal representations for struct types.


F21:
@zabetak Can this be merged? I'd like to get this into the next avatica release if possible.

zabetak:
I just waited somebody to have a second look! I am confident for the change and since @hsyuan had a look it should be ready to go ;)

PR #84, opened 2019-02-12
[CALCITE-2845] Avoid duplication of exception messages

vlsi:
https://issues.apache.org/jira/browse/CALCITE-2845


vlsi:
@joshelser , @risdenk , this PR passes all the tests, and it makes exceptions more readable.
I'm inclined to merge the change. Any thoughts / objections?

apilloud:
This change introduces a regression. Details of an exception are dropped in most JDBC clients, as they don't display the full stack trace. I've opened an issue: https://issues.apache.org/jira/browse/CALCITE-3040

vectorijk:
It blocks us for upgrading the version. Fix it in a minor version (1.14.1) release could work for us.
looking at the bug.

vlsi:
I would rather thread it as a SQL client bug.

PR #89, opened 2019-03-22
[CALCITE-2945] Use Boolean#equals in Boolean object compare

leesf:
Use Boolean#equals method in Boolean object compare in ConnectionPropertiesImpl#merge function.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/ConnectionPropertiesImpl.java
+++ b/core/src/main/java/org/apache/calcite/avatica/ConnectionPropertiesImpl.java
@@ -94,16 +94,16 @@ public class ConnectionPropertiesImpl implements Meta.ConnectionProperties {
     if (this == that) {
       return this;
     }
-    if (that.isAutoCommit() != null && !this.autoCommit.equals(that.isAutoCommit())) {
+    if (that.isAutoCommit() != null && !Objects.equals(this.autoCommit, that.isAutoCommit())) {
       this.autoCommit = that.isAutoCommit();
       this.isDirty = true;
     }
-    if (that.isReadOnly() != null && !this.readOnly.equals(that.isReadOnly())) {
+    if (that.isReadOnly() != null && !Objects.equals(this.readOnly, that.isReadOnly())) {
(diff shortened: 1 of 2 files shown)


leesf:
cc @risdenk @joshelser

risdenk:
FYI @leesf you don't need to cc myself or Josh on each PR. There are committers who see the JIRAs and PRs.

leesf:
@risdenk Sorry to bother you.

core/src/main/java/org/apache/calcite/avatica/ConnectionPropertiesImpl.java
```diff
@@ -94,11 +94,11 @@ public boolean isDirty() {
     if (this == that) {
       return this;
     }
-    if (that.isAutoCommit() != null && this.autoCommit != that.isAutoCommit()) {
+    if (that.isAutoCommit() != null && !this.autoCommit.equals(that.isAutoCommit())) {
       this.autoCommit = that.isAutoCommit();
       this.isDirty = true;
     }
-    if (that.isReadOnly() != null && this.readOnly != that.isReadOnly()) {
+    if (that.isReadOnly() != null && !this.readOnly.equals(that.isReadOnly())) {
```
danny0405: +1, nice catch.

danny0405:
@leesf It looks good, it would be better if you can supply a test case for the bad case when we use `==` as comparator.

leesf:
@danny0405 Updated this pr to address your comment.

julianhyde:
What about if that.readOnly is not null, but this.readOnly is null? The new code will throw NPE.

I suggest

```if (that.readOnly != null && !Objects.equals(this.readOnly, that.readOnly))```

The same bug exists in `autoCommit`.

We should use the same pattern in `transactionIsolation`, even though it doesn't have the bug.

In the test, please fix indentation, and use `assertThat( ... is())` rather than `assertEquals`.

leesf:
@julianhyde Thanks for your reviewing, updated this pr to address your comment.

PR #90, opened 2019-03-25
[CALCITE-2950] Avatica DriverVersion.load leaks InputStream

leonliao:
This commit fixes the leak problem at https://issues.apache.org/jira/browse/CALCITE-2950. The fix has been verified by using the procedure described in the JIRA issue. No memory leak observed after applying this fix.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/DriverVersion.java
+++ b/core/src/main/java/org/apache/calcite/avatica/DriverVersion.java
@@ -96,8 +96,9 @@ public class DriverVersion {
     int minorVersion = 0;
     int databaseMajorVersion = 0;
     int databaseMinorVersion = 0;
+    InputStream inStream = null;
     try {
-      final InputStream inStream =
+      inStream =
           driverClass.getClassLoader().getResourceAsStream(resourceName);
       if (inStream != null) {
         final Properties properties = new Properties();
@@ -138,6 +139,14 @@ public class DriverVersion {
       }
     } catch (IOException e) {
       e.printStackTrace();
+    } finally {
+      if (inStream != null) {
+        try {
+          inStream.close();
+        } catch (IOException ignore) {
+          ignore.printStackTrace();
+        }
+      }
     }
     return new DriverVersion(
         driverName, driverVersion, productName, productVersion,
--- a/core/src/main/java/org/apache/calcite/avatica/DriverVersion.java
+++ b/core/src/main/java/org/apache/calcite/avatica/DriverVersion.java
@@ -96,10 +96,8 @@ public class DriverVersion {
     int minorVersion = 0;
     int databaseMajorVersion = 0;
     int databaseMinorVersion = 0;
-    InputStream inStream = null;
-    try {
-      inStream =
-          driverClass.getClassLoader().getResourceAsStream(resourceName);
+    try (final InputStream inStream =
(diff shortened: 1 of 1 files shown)


leonliao:
The fix has been verified by using the procedure described in the JIRA issue. No memory leak observed after applying this fix.

danny0405:
Please add test cases for this PR.

joshelser:
> Please add test cases for this PR.

I'm not sure how we could add an effective test for this.

core/src/main/java/org/apache/calcite/avatica/DriverVersion.java
```diff
@@ -138,6 +139,14 @@ public static DriverVersion load(
       }
     } catch (IOException e) {
       e.printStackTrace();
+    } finally {
+      if (inStream != null) {
+        try {
+          inStream.close();
+        } catch (IOException ignore) {
+          ignore.printStackTrace();
```
joshelser: Could you re-throw the caught IOException as a `RuntimeException`? I can't think of a situation where this should ever fail. I'm not sure if we want this printed to stdout like you have it doing either (get a logger instance if you think this would be important. I think it would be fine to just re-throw this)

core/src/main/java/org/apache/calcite/avatica/DriverVersion.java
```diff
@@ -96,8 +96,9 @@ public static DriverVersion load(
     int minorVersion = 0;
     int databaseMajorVersion = 0;
     int databaseMinorVersion = 0;
+    InputStream inStream = null;
     try {
-      final InputStream inStream =
+      inStream =
```
vvysotskyi: Perhaps `try with resources` may be suitable here.

leonliao: Yes, you are right. I just checked https://calcite.apache.org/avatica/develop/avatica.html#download-source-build-and-run-tests, the required Java version is (JDK 1.8 or later, 1.9 preferred). Then try-with-resources is preferred here.

core/src/main/java/org/apache/calcite/avatica/DriverVersion.java
leonliao: For the test case, I mentioned the reproducing and verifying steps at https://github.com/leonliao/calcite-avatica-driver-version-test. And I didn't figure out a way to verify a native memory leak fix with just Java code. 

For the exception handling, after validating the minimum Java version is Java 8, I updated the code to use try-with-resources, then the previouslly existing IOException catch is reused.

joshelser:
It still gives me heartburn that the `catch (IOException)` at the bottom of this try-block doesn't re-throw the exception, but this isn't due to the change you've made. This change looks fine as-is.

@leonliao do you have interest in changing that next?

leonliao:
@joshelser 

The code implements a fallback logic if Exception happens, so exception should not be re-thrown?

joshelser:
> The code implements a fallback logic if Exception happens, so exception should not be re-thrown?

I'm more worried because I can't think of a "normal" reason that we should ever get an IOException thrown when reading a file like this. I would want to know that it happened via a hard-fail (unchecked exception).

julianhyde:
I'll go further. That kind of problem should be a java.lang.Error.

PR #91, opened 2019-03-29
[CALCITE-2961] Enable Travis to test against JDK 13

risdenk:


risdenk:
Hmmm JDK 13 fails due to what looks like an older ASM version. We should be able to override that.

risdenk:
This was fixed upstream in commit - https://github.com/spotbugs/spotbugs/commit/5153f67498394fcbbbe49f331229f24b74faf58b

No current release has that change yet.

risdenk:
Pushed a change that upgrades asm for Spotbugs.

risdenk:
Woot Travis passed with upgraded asm for spotbugs.

PR #94, opened 2019-04-10
[CALCITE-2939] NPE when executeBatch array type.

bakea:


core/src/main/java/org/apache/calcite/avatica/remote/TypedValue.java
```diff
@@ -359,6 +359,9 @@ private static Object serialToJdbc(ColumnMetaData.Rep type, ColumnMetaData.Rep c
           copy.add(serialToJdbc(componentRep, null, o, calendar));
         }
       }
+      if (componentRep == null && list.size() > 0) {
+        componentRep = ((TypedValue) list.get(0)).type;
+      }
```
danny0405: The document already say that the componentType in TypedValue is a non-null value [1], the only way is that when the nested parse in [2], which is a null we passed in initiatively.  For null i think it is reasonable cause we do not know what the value component type yet. The code you patched is reasonable if we finally got the list element type.

[1] https://github.com/apache/calcite-avatica/blob/63bab44362c808a8f4981e5d5f5af7d942982f3b/core/src/main/java/org/apache/calcite/avatica/remote/TypedValue.java#L165
[2] https://github.com/apache/calcite-avatica/blob/63bab44362c808a8f4981e5d5f5af7d942982f3b/core/src/main/java/org/apache/calcite/avatica/remote/TypedValue.java#L359

core/src/main/java/org/apache/calcite/avatica/remote/TypedValue.java
```diff
@@ -359,6 +359,9 @@ private static Object serialToJdbc(ColumnMetaData.Rep type, ColumnMetaData.Rep c
           copy.add(serialToJdbc(componentRep, null, o, calendar));
         }
       }
+      if (componentRep == null && list.size() > 0) {
+        componentRep = ((TypedValue) list.get(0)).type;
```
danny0405: If we still can not got the type, we should throw an exception, can you please add a check here ?

bakea: I think so.

F21:
@danny0405 Can you look at the updated changes?

PR #96, opened 2019-05-06
support from unix timestamp to string date

ambition119:
Calcite support sql function UNIX_TIMESTAMP and FROM_UNIXTIME, BuiltInMethod.java info like :
```java
UNIX_TIMESTAMP(DateTimeUtils.class, "unixTimestamp")

FROM_UNIXTIME(DateTimeUtils.class, "fromUnixTimestamp", long.class, TimeZone.class)
```

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java
@@ -16,6 +16,8 @@
  */
 package org.apache.calcite.avatica.util;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.text.DateFormat;
 import java.text.NumberFormat;
 import java.text.ParsePosition;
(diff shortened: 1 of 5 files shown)


julianhyde:
There are uses of default locale here. We can't have that. This code needs to behave identically in all locales.

In one or two places the comment says "unsigned int" when the result is clearly a signed long.

Methods that have a timezone argument must state very clearly, with examples, how the timezone is used. Are you converting to that timezone or from that timezone? So many times I expect a method to add 8 hours (to pacific) and it ends up subtracting 8 hours.

ambition119:
> There are uses of default locale here. We can't have that. This code needs to behave identically in all locales.
> 
> In one or two places the comment says "unsigned int" when the result is clearly a signed long.
> 
> Methods that have a timezone argument must state very clearly, with examples, how the timezone is used. Are you converting to that timezone or from that timezone? So many times I expect a method to add 8 hours (to pacific) and it ends up subtracting 8 hours.

Thank you for your suggestion, I will try to fix the problem caused by timezone.

julianhyde:
In SQL, timestamps are zoneless, whereas in Unix they are relative to the UTC epoch.

Existing methods that have 'unix' in the name and have `long` arguments and/or results follow the SQL semantics. (I apologize that the names are misleading. I wanted to capture the fact that it is milliseconds since 1970, but not that it was relative to 1970 UTC.)

Your method assumes that the `long` value is relative to UTC. That is opposite to how Avatica does things (although it is the natural way for Java folks to do things), and will confuse people.

ambition119:
> In SQL, timestamps are zoneless, whereas in Unix they are relative to the UTC epoch.
> 
> Existing methods that have 'unix' in the name and have `long` arguments and/or results follow the SQL semantics. (I apologize that the names are misleading. I wanted to capture the fact that it is milliseconds since 1970, but not that it was relative to 1970 UTC.)
> 
> Your method assumes that the `long` value is relative to UTC. That is opposite to how Avatica does things (although it is the natural way for Java folks to do things), and will confuse people.

thanks, Do you have better suggestions for implementation?

ambition119:
@julianhyde 
Currently I am a developer of the hadoop ecosystem, because we are most familiar with hive SQL, and there are already many Hive sql production tasks. Now we want to use Hive sql as the standard (so existing SQL tasks can be run without changes), running in Spark, Flink or Drill.
      Because calcite supports ANSI sql, many of them do not support well, such as [create funcation](https://github.com/apache/calcite/pull/908), [load date sql statement](https://issues.apache.org/jira/browse/CALCITE-2735) code not merge, and sql function like from_timestamp() [hive code is](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/udf/UDFFromUnixTime.java).
     At present, I have done a lot of work in the middle of the interception layer. I saw that Calcite has discussed the support of dialects like Hive. Can the submitted code be used?

julianhyde:
I'm open to including these changes in principle.

However, misunderstanding the difference between moments (e.g. millis since 1970 UTC) and zoneless timestamps (e.g. millis since zoneless 1970) is a major source of bugs. Especially as values flow between Java and SQL and back again. Avatica has to be 100% clear and consistent.

Your PR, as it stands, is confused on this issue.

ambition119:
> I'm open to including these changes in principle.
> 
> However, misunderstanding the difference between moments (e.g. millis since 1970 UTC) and zoneless timestamps (e.g. millis since zoneless 1970) is a major source of bugs. Especially as values flow between Java and SQL and back again. Avatica has to be 100% clear and consistent.
> 
> Your PR, as it stands, is confused on this issue.

My code is only available with a specific scope, thank you for your discussion and guidance.

PR #102, opened 2019-05-31
CALCITE-3104 Bump httpcore from 4.4.10 to 4.4.11

Fokko:
https://jira.apache.org/jira/browse/CALCITE-3104


Fokko:
@asereda-gs I've updated this commit message as well

PR #105, opened 2019-07-03
[CALCITE-3163] AbstractCursor#convertValue() now adheres to JDBC specification

ppanopticon:
Outlined in JIRA issue [CALCITE-3163]

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
@@ -1318,8 +1318,9 @@ public abstract class AbstractCursor implements Cursor {
         return componentAccessor.getInt();
       case Types.BIGINT:
         return componentAccessor.getLong();
-      case Types.FLOAT:
+      case Types.REAL:
         return componentAccessor.getFloat();
+      case Types.FLOAT:
       case Types.DOUBLE:
         return componentAccessor.getDouble();
       case Types.ARRAY:


core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -1318,8 +1318,9 @@ private Object convertValue() throws SQLException {
         return componentAccessor.getInt();
       case Types.BIGINT:
         return componentAccessor.getLong();
-      case Types.FLOAT:
+      case Types.REAL:
         return componentAccessor.getFloat();
+      case Types.FLOAT:
```
zabetak, line 1323: The fix looks reasonable but I was thinking that it would be nice to have some tests to guard against future regressions. Maybe in `ArrayTypeTest`?

asolimando: I agree, unfortunately `ArrayTypeTest` being based on `hsqldb`, it won't work as-is, because `hsqldb` maps `Double`, `Real` and `Float` to Java `double` (see [here](http://hsqldb.org/doc/guide/sqlgeneral-chapt.html#sgc_numeric_types)).

I tried and indeed a `CREATE TABLE $table (id integer, vals REAL ARRAY)` statement creates a table `(INTEGER, DOUBLE ARRAY)`.

Either we find an "easy" way to use another DB in those tests, or we need to find another location, I will have a look during the weekend.

asolimando: I have managed to add unit tests, I have opened a complementing PR: https://github.com/apache/calcite-avatica/pull/139

PR #107, opened 2019-07-08
CALCITE-3179: Bump Jackson from 2.9.8 to 2.9.9

Fokko:
To harmonize the versions of Jackson among the different projects.


Fokko:
@asereda-gs I've updated this commit message as well

PR #111, opened 2019-09-27
[CALCITE-3324] Add create method in MetaResultSet

rayokota:
Add an additional create method as suggested by @hsyuan in https://github.com/apache/calcite/pull/1441

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/Meta.java
+++ b/core/src/main/java/org/apache/calcite/avatica/Meta.java
@@ -577,6 +577,12 @@ public interface Meta {
           Objects.requireNonNull(signature), firstFrame, -1L);
     }
 
+    public static MetaResultSet create(String connectionId, int statementId,
+        boolean ownStatement, Signature signature, Frame firstFrame, long updateCount) {
+      return new MetaResultSet(connectionId, statementId, ownStatement,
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/calcite/avatica/Meta.java
```diff
@@ -577,6 +577,12 @@ public static MetaResultSet create(String connectionId, int statementId,
           Objects.requireNonNull(signature), firstFrame, -1L);
     }
 
+    public static MetaResultSet create(String connectionId, int statementId,
+        boolean ownStatement, Signature signature, Frame firstFrame, long updateCount) {
+      return new MetaResultSet(connectionId, statementId, ownStatement,
+          Objects.requireNonNull(signature), firstFrame, updateCount);
+    }
+
```
hsyuan: Hmm, this is OK. But why can't we make `MetaResultSet` constructor public? 
@julianhyde Do you know the reason?

julianhyde: If things are public it is easier for the user, harder for the maintainer. So my inclination is always to make things private, unless there is a good reason.

This is what I call a "drill a hole" change. Making something public without any rationale and without tests.

PR #119, opened 2020-01-17
[CALCITE-2704] Avoid use of ISO-8859-1 to parse request in JsonHandler

DonnyZone:
Credits to the [PR](https://github.com/apache/calcite-avatica/pull/85) from @vlsi.
Add a test for the fix.


joshelser:
Thanks for taking the time to test this, @DonnyZone. Let merge this.

PR #121, opened 2020-03-27
[CALCITE-3881][avatica] DateTimeUtils#addMonths yields incorrect resu…

docete:
…lts in some corner case

DateTimeUtils#addMonths use DateTimeUtils#ymdToUnixDate to calculate the JDN(julian day number). But in some corner cases it yields incorrent results. The root cause is: the algorithm of DateTimeUtils#ymdToUnixDate requires reasonable month(1 to 12)[1], but SqlFunctions#addMonths may pass in a month out of the reasonable range. This PR will fix it.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java
@@ -965,7 +965,20 @@ public class DateTimeUtils {
     return julian - EPOCH_JULIAN;
   }
 
+  /** Calculates the Julian Day Number for any date on the Gregorian. The paper of
+   * http://www.cs.utsa.edu/~cs1063/projects/Spring2011/Project1/jdn-explanation.html
+   * provides a brief explanation for this algorithm. */
   public static int ymdToJulian(int year, int month, int day) {
+  	if (month < 1 || month > 12) {
+  		throw new IllegalArgumentException(
+  			"Month of any date on the Gregorian should between 1 to 12, but " + month);
+    }
+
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java
```diff
@@ -965,7 +965,20 @@ public static int ymdToUnixDate(int year, int month, int day) {
...
   }
 
+  /** Calculates the Julian Day Number for any date on the Gregorian. The paper of
+   * http://www.cs.utsa.edu/~cs1063/projects/Spring2011/Project1/jdn-explanation.html
+   * provides a brief explanation for this algorithm. */
   public static int ymdToJulian(int year, int month, int day) {
+  	if (month < 1 || month > 12) {
+  		throw new IllegalArgumentException(
+  			"Month of any date on the Gregorian should between 1 to 12, but " + month);
+    }
```
danny0405: We should add some negative tests.

docete: negative tests for addMonths? 
See `checkAddMonths(2019, 9, 1, 2016, 8, 1, -37);`

danny0405: You should fix the throws of the test.

docete: Seems the throws come from old calcite since https://github.com/apache/calcite/pull/1890 is not merged.

PR #123, opened 2020-07-25
CALCITE-4138 Metadata operations via Avatica turn empty string args t…

stoty:
…o null

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/remote/Service.java
+++ b/core/src/main/java/org/apache/calcite/avatica/remote/Service.java
@@ -334,14 +334,19 @@ public interface Service {
         connectionId = msg.getConnectionId();
       }
 
+      //hasField() for String fields simply tells you if that field is ""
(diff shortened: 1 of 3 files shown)


joshelser:
This looks good to me. Protobuf3 has become such a bother with dropping the hasXXXXX methods. That would've made this easier.

Can you also make sure https://calcite.apache.org/avatica/docs/protobuf_reference.html is up to date with these changes, please? Source is at https://github.com/apache/calcite-site/tree/master/avatica

joshelser:
I'll pull this down and try to get it merged tonight/tmrw.

stoty:
Thanks @joshelser 
The docs PR is at https://github.com/apache/calcite-site/pull/2

I didn't check the history, but I think that the current code actually worked / would have worked with Protobuf2 .
Google has admitted to trading features for simplicity with v3. 

Even so, they should have just dropped the has* methods instead of returning false info.

PR #126, opened 2020-08-20
[CALCITE-4181] Fix exception when selecting List<Object>

kent-nguyen:
Currently Calcite supports array columns from both ArrayImpl or List input.

It also supports both struct and plain java object data format.

The issue is that an exception is thrown when data is a list of plain java objects, while it works fine when data is a list of struct or an array of plain java objects.

I have created unit tests to demonstrate the bug.

Currently these unit tests pass: arraysOfStructs(), listOfStructs(), arrayOfJavaObjects() but listOfJavaObject() is failed.

Method org.apache.calcite.avatica.util.AbstractCursor.ArrayAccessor.convertValue() misses a case for Types.OTHER, which is for general java objects, hence the exception is thrown.

This pull request adds that missing type and makes all four unit tests pass. 

Reference link: https://github.com/apache/calcite-avatica/blob/master/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java#L1346

### DIFF
--- a/core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
+++ b/core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
@@ -33,6 +33,7 @@ import java.sql.Types;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
```diff
@@ -164,6 +165,267 @@
     }
   }
 
+  /**
+   * The same test as arrayOfStructs(), except we use List instead of ArrayImpl.
+   *
+   * This test deliberately covers the
```
asolimando, line 171: Nitpick: I think the first line of the comment is enough, if you feel more context is needed, you can also refer to the JIRA ticket.

core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
```diff
@@ -164,6 +165,267 @@
...
+   * org.apache.calcite.avatica.util.AbstractCursor.ArrayAccessor.convertValue()
+   * method, which is called from
+   * org.apache.calcite.avatica.util.AbstractCursor.ArrayAccessor.getObject() when
+   * input is a general List instead of org.apache.calcite.avatica.util.ArrayImpl.
+   *
+   * @throws Exception
+   */
+  @Test
+  public void listOfStructs() throws Exception {
+    // Define the struct type we're creating.
```
asolimando, line 181: Nitpick: I'd remove the comments like "creating/defining x, y", it is clear from the code itself in most cases

core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
```diff
@@ -164,6 +165,267 @@
...
+    ColumnMetaData intMetaData = MetaImpl.columnMetaData("MY_INT", 1, int.class, false);
+    ColumnMetaData stringMetaData = MetaImpl.columnMetaData("MY_STRING", 2, String.class, true);
+    StructType structType = ColumnMetaData.struct(Arrays.asList(intMetaData, stringMetaData));
+    // Create some structs.
+    Struct struct1 = new StructImpl(Arrays.<Object>asList(1, "one"));
```
asolimando, line 186: The diamond operator can be omitted for `Arrays.<>` and `Collections.<>` function calls (there are many occurrences in the whole file)

core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
```diff
@@ -164,6 +165,267 @@
...
+    Struct struct2 = new StructImpl(Arrays.<Object>asList(2, "two"));
+    Struct struct3 = new StructImpl(Arrays.<Object>asList(3));
```
asolimando, line 188: Nitpick: try to be consistent in the use of `Collections.singletonList` vs `Arrays.asList(x)` overall in the file

core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
```diff
@@ -164,6 +165,267 @@
...
+    Struct struct4 = new StructImpl(Arrays.<Object>asList(4, "four"));
+    ArrayType arrayType = ColumnMetaData.array(structType, "OBJECT", Rep.STRUCT);
+    ColumnMetaData arrayMetaData = MetaImpl.columnMetaData("MY_ARRAY", 1, arrayType, false);
+    ArrayImpl.Factory factory = new ArrayFactoryImpl(Unsafe.localCalendar().getTimeZone());
+
+    // Create some lists from the structs, instead of an array.
+    List<Object> list1 = new ArrayList<Object>(2);
```
asolimando, line 195: Nitpick: why not using `Arrays.asList(struct1, struct2);` as you do in the other cases?

core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
```diff
@@ -164,6 +165,267 @@
...
+      assertEquals(2, o.length);
+      assertEquals(4, o[0]);
+      assertEquals("four", o[1]);
+    }
+  }
+
+  /**
+   * Plain Java object class for the two tests that follow.
+   */
+  class PlainJavaObject {
```
asolimando, line 248: The class can be `static`

core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
```diff
@@ -164,6 +165,267 @@
...
+      actualStruct = (Struct) arrayData[1];
+      o = actualStruct.getAttributes();
+      assertEquals(2, o.length);
+      assertEquals(4, o[0]);
+      assertEquals("four", o[1]);
+    }
+  }
+
+  /**
+   * Plain Java object class for the two tests that follow.
```
asolimando, line 246: The comment does not seem to add much value, I think it can be safely removed

asolimando (commented):
It looks overall good to me, I have left some small comments, some address issues occurring multiple times, I have noted them mostly once, please make sure to address them wherever it applies.

core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
kent-nguyen: :core:checkstyleTest would fail with error "Missing a Javadoc comment" if we remove that docblock.

kent-nguyen:
Thanks very much for your review @asolimando, really appreciate your guidance. I've made the changes as suggested, please let me know if there's anything else I should address.

core/src/test/java/org/apache/calcite/avatica/util/ArrayImplTest.java
```diff
@@ -164,6 +165,267 @@
...
+      actualObject = (PlainJavaObject) arrayData[1];
+      assertEquals(4, actualObject.intProperty);
+      assertEquals("four", actualObject.stringProperty);
+    }
+  }
+
+  /**
+   * Test case for when a column is a list of plain Java objects.
+   *
+   * @throws Exception
```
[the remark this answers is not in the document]
asolimando: My suggestion here was to turn from the current comment to a plain one along this line (since you are not filling the "javadoc" part, it looks like you want just to add a multi-line comment):
```java
  /*
   * Test case for when a column is an array of plain Java objects.
   * This is a common use case when data come from a dynamic schema source.
   */
```

asolimando (changes requested):
I have just a single pending comment on the javadoc vs multi-line (plain) comment, then it's good to me.

Unfortunately my acceptance won't be enough to merge the PR since I am not a committer, but usually a review helps moving things forward anyway :)

kent-nguyen:
Thanks very much for your feedback @asolimando.
I have made the change as your last comment.
Hopefully with your acceptance, it is gradually got merged.

rubenada:
Thanks for the contribution @kent-nguyen 
Please bear in mind that, as per our [contribution guidelines](https://calcite.apache.org/develop/#contributing):
- Commits must be squashed into a single one
- The commit message should describe the problem (typically they should not start with "Fix..."), and also it must finish with your name in parenthesis. Maybe a better alternative in this case would be using the Jira title as commit message: "Avatica throw exception when select field is a List<Object> (Kent Nguyen)"

kent-nguyen:
Thank you @rubenada for your guideline.
I have squashed the commits, and updated the message in instructed.
Please let me know if there is anything else I should address.

asolimando (approved):
LGTM, thanks for the updates @kent-nguyen

rubenada:
I am not too familiar with avatica module, but the change LGTM; and it provides the required unit tests.
I will merge it in the next 24/48 h if nobody presents any objection.

@kent-nguyen could you please review a minor deatil in the commit message (sorry you probably copied-pasted it from my message, which was wrong and I edited it afterwards)?
`Avatica throws exception when select field is a List (Kent Nguyen)` 
should be
`Avatica throws exception when select field is a List<Object> (Kent Nguyen)`

kent-nguyen:
Thank you @rubenada 
I have updated the commit message on that detail.
I am looking forward to the merge. Cheers.

PR #127, opened 2020-08-26
CALCITE-4196 Consume all data from client before replying with HTTP…

joshelser:
…/401

SPNEGO's handshake involves sending an HTTP/401 to "challenge" the
client to reply with authentication data. If the client is sending
a significant amount of data in the original request, the client
will still be writing this data when the server replies. This causes
the client to receive a TCP Reset when it continues to write data, and
ultimately manifests in a "Broken Pipe" runtime exception.

The fix is to simply consume all data the client wrote prior to
responding with the HTTP/401.

This also changed some stuff in AvaticaProtobufHandler to reduce a level of nested conditionals and fix a Logger class. The overall flow/purpose of that class did not change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,3 @@ settings.xml
 /target/
 /*/target/
 /shaded/*/target/
-bin
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ settings.xml
 /target/
 /*/target/
 /shaded/*/target/
+bin
--- a/server/src/main/java/org/apache/calcite/avatica/server/AbstractAvaticaHandler.java
+++ b/server/src/main/java/org/apache/calcite/avatica/server/AbstractAvaticaHandler.java
@@ -23,8 +23,6 @@ import org.apache.calcite.avatica.remote.Service.ErrorResponse;
 
 import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.handler.AbstractHandler;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.HttpURLConnection;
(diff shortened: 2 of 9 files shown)


server/src/main/java/org/apache/calcite/avatica/server/AbstractAvaticaHandler.java
```diff
@@ -34,6 +38,7 @@
  */
 public abstract class AbstractAvaticaHandler extends AbstractHandler
     implements MetricsAwareAvaticaHandler {
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractAvaticaHandler.class);
```
risdenk, line 41: nit: Looks like LOG is potentially unused?

joshelser:
Forgot to mention -- I tried to write a unit test which demonstrates the 401 and the broken pipe, but I wasn't able to. My only guess is that going against localhost/loopback instead of over the network was avoiding the issue somehow. Didn't seem worth it to figure out why that didn't work when I had reproduced the issue with Apache Phoenix and was able to validate that this fix works.

server/src/main/java/org/apache/calcite/avatica/server/AbstractAvaticaHandler.java
joshelser: ah, yup. I had some logging in here as a part of my work, but considered none of it beneficial long-term.

.gitignore
```diff
@@ -22,3 +22,4 @@ settings.xml
 /target/
 /*/target/
 /shaded/*/target/
+bin
```
vlsi, line 25: This would ignore all bin folders in all subfolders (e.g. it would ignore `org.apache.calcite.encodings.bin....` folder if one is created). Do you intend that?

joshelser: Hrm, I did not. I don't think we have such folders in avatica, but that's not great either. Let me revert that.

Thanks @vlsi

PR #130, opened 2020-11-05
[CALCITE-4379] Meta.Frame created with java float values in rows hits a ClassCastException in toProto()

dimas-b:
* Remove unnecessary conversion of Float values to long in
  TypedValue.toProto(...). The subsequently invoked writeToProtoWithType(...)
  method actually expects float values in this case.

* Make a similar fix for Character values.

* Add Frame serialization round-trip unit tests to cover this failure mode.

* Note: a similar problem appears to still exist for Timestamp and Date/Time
  values in this context, but I did not attempt to fix it in this commit because
  there is apparently no use case for putting Timestamp/Date/Time objects into
  Frames. Those values are normally returned as plain numbers in result rows.

### DIFF
--- a/core/src/test/java/org/apache/calcite/avatica/FrameTest.java
+++ b/core/src/test/java/org/apache/calcite/avatica/FrameTest.java
@@ -23,6 +23,7 @@ import org.apache.calcite.avatica.proto.Common.TypedValue;
 
 import org.junit.Test;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


asolimando (approved):
The PR LGTM, regarding the `Timestamp`/`Date` conversion to `long` I'd expect the roundtrip`fromProto(toProto)` to be always equivalent to its input, independently from the specific type; but since I don't know either the rationale for having that conversion in the first place, I agree we can leave it out from the current PR, unless someone can shed more light on it and we come to an agreement on how to deal with it.

core/src/test/java/org/apache/calcite/avatica/FrameTest.java
```diff
@@ -105,6 +106,84 @@ public void testMultipleRows() {
     serializeAndTestEquality(singleRow);
   }
 
+  protected void testValueRoundTrip(Object value) {
```
joshelser, line 109: Can you please update `testSingleRow` in this class to also use this helper method you've created?

joshelser (changes requested):
LGTM -- essentially, we were trying to double-convert, it looks like? `writeToProtoWithType` was already trying to handle the serialization into the protobuf.

One ask to make the test file you modified consistent (one method is not using the nice helper you wrote).

I assume the answer is "you don't know", but is there an issue on the JSON parsing side?

dimas-b:
> but is there an issue on the JSON parsing side?

As fare as I can tell this particular problem does not exist on the JSON serialization path. Apparently `Meta.Frame` fields are accessed directly in that case and its `toProto()` method is not called.

core/src/test/java/org/apache/calcite/avatica/FrameTest.java
dimas-b: Sorry, @joshelser, I'm not sure what you mean.

`testSingleRow`, unlike `testValueRoundTrip`, uses a multi-value row and both call the existing `serializeAndTestEquality` method to do actual validation.

`testValueRoundTrip` is essentially a convenience method for deal with single-valued rows.

joshelser: Ah, sorry. I see. I was getting "one row" mixed up with "one value in that row". 👍

joshelser (approved):
LGTM, thanks for your comment about the JSON path. I'll try to get this merged today.

PR #131, opened 2020-12-24
[CALCITE-4367] Correct Avatica protocol docs

joshelser:
isDirty needs to be deprecated and removed as it should
not have been in the protocol to begin with.


john-bodley (approved):

risdenk (approved):

joshelser:
Thanks folks!

PR #132, opened 2021-01-01
[CALCITE-4152] Switch to ConfigurableSpnego Jetty implementations

joshelser:
The new ConfigurableSpnego* classes in Jetty give us the ability to use the SessionHandler to send back a session cookie and avoid having to execute the spnego handshake for each and every call. This is a big improvement in reducing the traffic to the avatica server.

*Still a draft*


joshelser:
Just updated this pull request. Getting closer, but still not there.

YCSB test results were _very_ promising. In a nutshell, doing `ycsb load` against Avatica with hsqldb showed roughly 2x performance improvement (execution time) with a batch size of 5 rows per `executeUpdate`. This is because we're actually making about half the number of HTTP calls we were previously making. Similarly, with a large batch size (1000rows/batch), execution time is roughly equivalent with the previous performance. The Java client looks great.

However, when I tried to use the Python client with Spnego, I ran into issues. Specifically, I was seeing a NullPointerException where the result from `GSSContext.acceptSecContext` (inside Avatica) was returning null. Sadly, the JVM docs don't tell me why this might be and I wasn't able to dig into the JVM source enough to peel back the layers of indirection. As such, this still isn't ready to go.

It would be great to get a set of eyes on this now as there is some breaking runtime semantics.
1. "additionalAllowedRealms" is no more. If you're authenticated via Kerberos, you're allowed in.
2. There is no external Kerberos authentication for the Avatica server allowed anymore. Jetty will always log in as the given principal/keytab.

PR #134, opened 2021-01-23
[CALCITE-4476] DateTimeUtils.timeStringToUnixDate may produce wrong time (Vladimir Ozerov)

devozerov:
Fixes [#4476](https://issues.apache.org/jira/browse/CALCITE-4476)


rubenada:
Thanks for the fix @devozerov 
Please note that, per our [contributing rules](https://calcite.apache.org/develop/#contributing):
- Commit message should describe the bug and they should not start with "Fix...". In this case, I think the title you chose for the Jira would be a better message "DateTimeUtils.timeStringToUnixDate may produce wrong time".
- If you are not a committer, add your name in parentheses at the end of the message.

devozerov:
Hi @rubenada, sorry, I've missed the guidelines. I've changed the message as you suggested.

PR #138, opened 2021-02-15
[CALCITE-2489] Order of fields in JavaTypeFactoryImpl#createStructTyp…

asolimando:
…e is unstable

The changes in the PR are meant to allow us to "pass" the field ordering and avoid any call to Class.getFields(), which is problematic as it's JVM dependent.

More details are available in the [ML discussion](https://lists.apache.org/thread.html/r7537572f56fe2d734baad2de7fdeda490deb520febc3072c60e79de0%40%3Cdev.calcite.apache.org%3E)

EDIT: let me provide a more detailed description/overview to ease the reviewing process.

The PR introduces a deprecation annotation for method 
```
public RecordIteratorCursor(Iterator<E> iterator, Class<E> clazz) {
    this(iterator, clazz, Arrays.asList(clazz.getFields()));
}
```
This is necessary to adopt a Calcite-specific and deterministic/stable fields ordering.
This is possible by passing the fields (in the sought order) via `RecordIteratorCursor(Iterator, Class, List)`, as already done for record with projected columns (in order to know what to keep/drop a List of fields is explicitly passed).

In order to use the same method for generic records (projected and non-projected), `Meta.java` had to be updated, to make some methods more generic (they were "hard-coding"  `Style.RECORD_PROJECTION` as it was the only expected user of those methods, while now we want a `Style` parameter to accommodate both `Style.RECORD_PROJECTION` and `Style.RECORD`).

The only "glitch" is that `MetaImpl` internal classes (to expose different metadata, the catalog etc.) have columns not matching the field names (they are in capital letters, etc.).
On Calcite side it suffices to pass `null` instead of the class name for those classes (it's a single call in Calcite, I have already tried that locally), so the code won't try to validate the column names against the class fields (only done when a class is provided, it's a "sanity check", but the precondition is to have obtained the column names from the class fields, which is not the case for those internal classes).

So, as per the comment in the code, once we merge `CALCITE-2489` in Calcite codebase we can drop the ugly `if` statement.

Changes in `MetaImpl.java`, are simply taking advantage of the more general signature of some methods in `Meta.java`, where the code had to differentiate between `Style.RECORD_PROJECTION` and `Style.RECORD`, now we can call a single method.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/Meta.java
+++ b/core/src/main/java/org/apache/calcite/avatica/Meta.java
@@ -41,6 +41,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.stream.Collectors;
 
 /**
  * Command handler for getting various metadata. Should be implemented by each
@@ -608,8 +609,8 @@ public interface Meta {
     private CursorFactory(Style style, Class clazz, List<Field> fields,
         List<String> fieldNames) {
       assert (fieldNames != null)
-          == (style == Style.RECORD_PROJECTION || style == Style.MAP);
-      assert (fields != null) == (style == Style.RECORD_PROJECTION);
+          == (style == Style.RECORD || style == Style.RECORD_PROJECTION || style == Style.MAP);
+      assert (fields != null) == (style == Style.RECORD || style == Style.RECORD_PROJECTION);
       this.style = Objects.requireNonNull(style);
       this.clazz = clazz;
       this.fields = fields;
@@ -628,9 +629,9 @@ public interface Meta {
       case LIST:
         return LIST;
       case RECORD:
-        return record(clazz);
+        return record(Style.RECORD, clazz, null, fieldNames);
       case RECORD_PROJECTION:
-        return record(clazz, null, fieldNames);
+        return record(Style.RECORD_PROJECTION, clazz, null, fieldNames);
       case MAP:
         return map(fieldNames);
       default:
@@ -647,12 +648,13 @@ public interface Meta {
     public static final CursorFactory LIST =
         new CursorFactory(Style.LIST, null, null, null);
 
-    public static CursorFactory record(Class resultClazz) {
-      return new CursorFactory(Style.RECORD, resultClazz, null, null);
+    public static CursorFactory record(Class resultClass, List<Field> fields,
+                                       List<String> fieldNames) {
+      return record(Style.RECORD_PROJECTION, resultClass, fields, fieldNames);
     }
 
-    public static CursorFactory record(Class resultClass, List<Field> fields,
-        List<String> fieldNames) {
+    public static CursorFactory record(Style style, Class resultClass, List<Field> fields,
+                                       List<String> fieldNames) {
       if (fields == null) {
         fields = new ArrayList<>();
         for (String fieldName : fieldNames) {
@@ -663,8 +665,7 @@ public interface Meta {
           }
         }
       }
-      return new CursorFactory(Style.RECORD_PROJECTION, resultClass, fields,
-          fieldNames);
+      return new CursorFactory(style, resultClass, fields, fieldNames);
     }
 
     public static CursorFactory map(List<String> fieldNames) {
@@ -685,7 +686,14 @@ public interface Meta {
       if (List.class.isAssignableFrom(resultClazz)) {
         return LIST;
       }
-      return record(resultClazz);
+      // columns for MetaImpl internal classes do not match field names, "record" would fail
+      // can be removed once CALCITE-2489 is merged in Calcite
+      if (resultClazz.getName().startsWith("org.apache.calcite.avatica.MetaImpl")
+          && !resultClazz.getName().endsWith("MetaTable")) {
+        return ARRAY;
+      }
+      return record(Style.RECORD, resultClazz, null,
+          columns.stream().map(c -> c.columnName).collect(Collectors.toList()));
     }
 
     public Common.CursorFactory toProto() {
--- a/core/src/main/java/org/apache/calcite/avatica/Meta.java
+++ b/core/src/main/java/org/apache/calcite/avatica/Meta.java
@@ -672,6 +672,16 @@ public interface Meta {
       return new CursorFactory(Style.MAP, null, null, fieldNames);
     }
 
+    /**
+     * Deduces the appropriate {@code CursorFactory} for accessing the underlying
+     * result set. For result sets composed by records, {@code resultClazz} must
+     * be not null, and each field name in {@code columns} must match one of its
+     * public fields.
+     * @param columns The columns metadata for the result set
+     * @param resultClazz The class representing the records, if any
+     * @return the appropriate {@code CursorFactory} for the underlying result set
+     * @throws RuntimeException if the field name validation fails
+     */
     public static CursorFactory deduce(List<ColumnMetaData> columns,
         Class resultClazz) {
       if (columns.size() == 1) {
--- a/core/src/main/java/org/apache/calcite/avatica/MetaImpl.java
+++ b/core/src/main/java/org/apache/calcite/avatica/MetaImpl.java
@@ -89,14 +89,9 @@ public abstract class MetaImpl implements Meta {
           (Iterable<Object[]>) (Iterable) iterable;
       return new ArrayIteratorCursor(iterable1.iterator());
     case RECORD:
-      @SuppressWarnings("unchecked") final Class<Object> clazz =
-          cursorFactory.clazz;
-      return new RecordIteratorCursor<Object>(iterable.iterator(), clazz);
     case RECORD_PROJECTION:
-      @SuppressWarnings("unchecked") final Class<Object> clazz2 =
-          cursorFactory.clazz;
-      return new RecordIteratorCursor<Object>(iterable.iterator(), clazz2,
-          cursorFactory.fields);
+      @SuppressWarnings("unchecked") final Class<Object> clazz = cursorFactory.clazz;
+      return new RecordIteratorCursor<>(iterable.iterator(), clazz, cursorFactory.fields);
     case LIST:
       @SuppressWarnings("unchecked") final Iterable<List<Object>> iterable2 =
           (Iterable<List<Object>>) (Iterable) iterable;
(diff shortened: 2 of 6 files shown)


F21:
Discussion for this change is available on the mailing list here: https://lists.apache.org/thread.html/r7537572f56fe2d734baad2de7fdeda490deb520febc3072c60e79de0%40%3Cdev.calcite.apache.org%3E

core/src/main/java/org/apache/calcite/avatica/Meta.java
```diff
@@ -647,11 +648,12 @@ public static CursorFactory create(@JsonProperty("style") Style style,
     public static final CursorFactory LIST =
         new CursorFactory(Style.LIST, null, null, null);
 
-    public static CursorFactory record(Class resultClazz) {
-      return new CursorFactory(Style.RECORD, resultClazz, null, null);
+    public static CursorFactory record(Class resultClass, List<Field> fields,
+                                       List<String> fieldNames) {
+      return record(Style.RECORD_PROJECTION, resultClass, fields, fieldNames);
```
julianhyde, line 653: Fix the indentation of line 652, `List<String> fieldNames) {`. Then it will be clear in the diff that this method's signature has not changed.

asolimando: It's done, sorry to have missed that.

asolimando:
I realized that there were no tests in Avatica specifically covering the lines changed by the PR.
Even if the changed code is validated indirectly by the Calcite tests, I felt it was better to have unit-tests in Avatica too for that, the last commit adds them.

asolimando:
Rebased on current master

F21:
ping @julianhyde

julianhyde:
Sorry, no, I'm not planning to commit this. I comment on a lot of changes; that doesn't automatically make me the committer.

core/src/main/java/org/apache/calcite/avatica/Meta.java
```diff
@@ -647,12 +648,13 @@ public static CursorFactory create(@JsonProperty("style") Style style,
     public static final CursorFactory LIST =
         new CursorFactory(Style.LIST, null, null, null);
 
-    public static CursorFactory record(Class resultClazz) {
-      return new CursorFactory(Style.RECORD, resultClazz, null, null);
```
zabetak, line 651: Maybe we should keep the method for a short while for backwards compatibility. WDYT?

core/src/main/java/org/apache/calcite/avatica/MetaImpl.java
```diff
@@ -139,15 +134,8 @@ protected Getter createGetter(int ordinal) {
       return list;
     case RECORD:
     case RECORD_PROJECTION:
```
zabetak, line 136: With the current PR the behavior of `RECORD` and `RECORD_PROJECTION` seems to be exactly the same. Do we need to keep both?

zabetak (commented):
Thanks for the PR @asolimando ! Left a few small comments and tried to resolve them at the same time in https://github.com/zabetak/calcite-avatica/tree/CALCITE-4503. Can you check the changes and let me know what you think.

core/src/main/java/org/apache/calcite/avatica/Meta.java
```diff
@@ -685,7 +686,14 @@ public static CursorFactory deduce(List<ColumnMetaData> columns,
       if (List.class.isAssignableFrom(resultClazz)) {
         return LIST;
       }
-      return record(resultClazz);
+      // columns for MetaImpl internal classes do not match field names, "record" would fail
+      // can be removed once CALCITE-2489 is merged in Calcite
+      if (resultClazz.getName().startsWith("org.apache.calcite.avatica.MetaImpl")
+          && !resultClazz.getName().endsWith("MetaTable")) {
+        return ARRAY;
+      }
```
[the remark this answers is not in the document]
rubenada: Also, probably not our concern, but just for the record: this piece of code might not work in a downstream project if shading is involved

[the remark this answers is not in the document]
zabetak: With or without the if statement the `deduce` method changes behavior and it might break downstream projects (as it breaks Calcite for instance). I think it is worth adding a backward compatibility note for the method and the change in general in `history.md`.

core/src/main/java/org/apache/calcite/avatica/MetaImpl.java
asolimando: I agree, we can unify them (LGTM in your branch)

core/src/main/java/org/apache/calcite/avatica/Meta.java
asolimando: Agreed, not sure if we should target `2.0` like for the other "deprecated" annotations I have seen around, rather than `1.19` as I have seen in your branch. Both work for me.

core/src/main/java/org/apache/calcite/avatica/Meta.java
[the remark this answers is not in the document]
asolimando: I was not sure if such a "breaking change" was acceptable, in that case I agree we can let Calcite adapt to the change. This allows us to avoid the shading issue @rubenada mentions, thanks for the suggestion btw.

For `history.md` a note along these lines could work?

> `Meta#deduce(List<ColumnMetaData> columns, Class resultClazz)` now derives the order of the fields from the list of provided column metadata `columns` when generating a record from the given Java class `resultClazz`, instead of relying on the field order provided by `Object#getFields()`, which is a JVM-dependent feature.

`history.md` seems to be updated during the release itself by combining the list of commits and associated comments. I don't know what's the usual way for adding such a note, maybe it can be part of the commit message itself?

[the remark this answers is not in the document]
zabetak: Usually when there is a breaking change people update `history.md` directly when they commit the change. During the release the release manager will make a pass on `history.md` to add commits etc. and reorganise/reword the content as they see fit.

The message looks good although it doesn't explain for instance that previous working code might now throw something like the following:  `java.lang.RuntimeException: java.lang.NoSuchFieldException: TABLE_CAT` Why this happens and how should they fix it?

core/src/main/java/org/apache/calcite/avatica/Meta.java
zabetak: I am not sure if there we are ever going to have a `2.0` :) As you prefer, I dont mind.

zabetak:
@asolimando Regarding the change that is neccessary to do in Calcite for the tests to pass do we need to wait for CALCITE-2489 or it can be done independently?

core/src/main/java/org/apache/calcite/avatica/Meta.java
asolimando: Let's leave 1.19 then, by the next release we will have time to hear if there is any concerns from downstream projects.

asolimando:
> @asolimando Regarding the change that is neccessary to do in Calcite for the tests to pass do we need to wait for [CALCITE-2489](https://issues.apache.org/jira/browse/CALCITE-2489) or it can be done independently?

The fix does not depend on CALCITE-2489 and can be applied independently, it just boils down to replacing `clazz` with `null` at this line in Calcite: 
https://github.com/apache/calcite/blob/4bc916619fd286b2c0cc4d5c653c96a68801d74e/core/src/main/java/org/apache/calcite/jdbc/CalciteMetaImpl.java#L200

I can file a JIRA ticket in Calcite for that and prepare a PR so we are ready for Avatica 1.18.

core/src/main/java/org/apache/calcite/avatica/Meta.java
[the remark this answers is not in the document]
asolimando: Thanks @zabetak, I have added a new commit with an update to `history.md` with the explanation of the change. I haven't included the changes from your branch, although they all LGTM.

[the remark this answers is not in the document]
asolimando: I have also added javadoc to `deduce` to cover explicitly the expectations for columns and resultclazz, and that an exception is thrown on validation failure.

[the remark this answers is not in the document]
zabetak: Adding javadoc is a very good idea, thanks @asolimando !

zabetak:
> I can file a JIRA ticket in Calcite for that and prepare a PR so we are ready for Avatica 1.18.

Great let's do that and I will take it from there (rebase and squash everything along with the extra commits and push to master).

asolimando:
> > I can file a JIRA ticket in Calcite for that and prepare a PR so we are ready for Avatica 1.18.
> 
> Great let's do that and I will take it from there (rebase and squash everything along with the extra commits and push to master).

Thanks a lot @zabetak, much appreciated. I have filed [CALCITE-4556](https://issues.apache.org/jira/browse/CALCITE-4556) and associated PR which is already passing the tests in Calcite, it does not even need to wait for the fix on Avatica's side.

zabetak:
@asolimando it seems there are more failures in Calcite: https://github.com/zabetak/calcite-avatica/runs/2216366873?check_suite_focus=true#step:5:468

Can you please have a look?

asolimando:
> @asolimando it seems there are more failures in Calcite: https://github.com/zabetak/calcite-avatica/runs/2216366873?check_suite_focus=true#step:5:468
> 
> Can you please have a look?

That's weird, I will have a look asap, it looks like an array-based cursor is used even for non-empty metadata result sets, instead of a record-based one, maybe one of the latest changes is causing that as a side-effect.

Btw, I am having issues lately to use the sources from my local `calcite-avatica` project in Calcite in IntelliJ.

I usually uncomment in `gradle.properties` the line `localAvatica=../calcite-avatica`, Avatica is correctly imported into IntelliJ, and it's the version that is used. Now, the importing bit works, but when it comes to debug the unit-tests, it's using 1.27.0 in the local maven cache. I did not have this problem before, I don't know where this is coming from, am I the only one experiencing that?

asolimando:
Hi @zabetak, I have fixed the issue and opened a PR against your branch, you can find it here: https://github.com/zabetak/calcite-avatica/pull/1

PR #141, opened 2021-04-09
[CALCITE-4573] Fix  close AvaticaResultSet  will throw NPE bug

chenyuzhi459:
Fix bug [\[CALCITE-4573\]](https://issues.apache.org/jira/browse/CALCITE-4573)
## Problem:
In `ArrayFactoryImpl.create(AvaticaType elementType, Iterable<Object> elements)` method, we will create AvaticaResultSet with a null statement, However in the `close()` menthod  for AvaticaResultSet, it doesn't valid null situation before we call ` statement.onResultSetClose(this);`, which cause npe exception.

## Solution:
add check for statement in  `close()` method

## Key changed/added classes in this PR   
- `org.apache.calcite.avatica.AvaticaResultSet`

### DIFF
--- a/core/src/test/java/org/apache/calcite/avatica/AvaticaClosedResultSetWithNullStatmentTest.java
+++ b/core/src/test/java/org/apache/calcite/avatica/AvaticaClosedResultSetWithNullStatmentTest.java
@@ -36,7 +36,7 @@ import static org.junit.Assert.assertTrue;
 @RunWith(Parameterized.class)
 public class AvaticaClosedResultSetWithNullStatmentTest extends AvaticaClosedTestBase<ResultSet> {
   // Mapping between Connection method and the verifier to check close behavior
-  private static final Function<Method, MethodVerifier> METHOD_MAPPING = method -> {
+  private static MethodVerifier methodVerifier(Method method) {
     String name = method.getName();
     // All update methods are not supported yet
     if (name.startsWith("update")) {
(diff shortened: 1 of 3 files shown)


rubenada:
Thanks for the fix. It seems there are code style violations.
Is there a Jira ticket for this issue?

chenyuzhi459:
> Thanks for the fix. It seems there are code style violations.
> Is there a Jira ticket for this issue?

Thanks for remindig me, I have fixed the code style violations using `./gradlew autostyleApply`.  I'd like to know how to create a  Jira ticket for this issue, beause it's I have no expirence for this yet?

core/src/test/java/org/apache/calcite/avatica/AvaticaClosedResultSetWithNullStatmentTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests for {@code AvaticaResultSet} relative to close behavior
+ */
+@RunWith(Parameterized.class)
+public class AvaticaClosedResultSetWithNullStatmentTest extends AvaticaClosedTestBase<ResultSet> {
+  // Mapping between Connection method and the verifier to check close behavior
+  private static final Function<Method, MethodVerifier> METHOD_MAPPING = method -> {
```
vlsi, line 39: Please use the regular Java method instead of Function field.

For instance: `private static MethodVerifier methodVerifier(Method method)`

chenyuzhi459: Thanks for your tips, I have chaned it to a method

PR #144, opened 2021-05-11
[CALCITE-4575] Remove Gradle wrapper from source distribution

julianhyde:
Please review, but do not merge.

@F21, I think you should merge this (with corrections, if necessary) during your process, as release manager, to create 1.18.0 RC1.


julianhyde:
I have added a commit with a fix for https://issues.apache.org/jira/browse/CALCITE-4576. Please review that also.

F21:
Was the fix for CALCITE-4576 a separate commit? I think you forgot to push it, so you will need to open a new PR for it.

julianhyde:
I've pushed it to the same branch. No PR needed.

PR #145, opened 2021-05-12
[CALCITE-4605] Unnecessary Prefix assert in UnregisteredDriver#connect

wangxianghu:
The URL prefix has already been verified in `org.apache.calcite.avatica.UnregisteredDriver#acceptsURL`, so the following assertion is unnecessary.


wangxianghu:
Not a problem

PR #147, opened 2021-06-12
CALCITE-4646 Update to Jetty 9.4.42.v20210604

stoty:


stoty:
The failing Windows test may be caused by the new Jetty version.
This patch doesn't touch anything in the Avatica codebase that could case that.

stoty:
The test fixes are for removing an illegal leading slash in the file paths on Windows.
These were accepted by earlier Jetty versions, but not by 9.4.42.

server/src/test/java/org/apache/calcite/avatica/server/BasicAuthHttpServerTest.java
```diff
@@ -44,8 +44,8 @@
   private static String url;
 
   @BeforeClass public static void startServer() throws Exception {
-    final String userPropertiesFile = URLDecoder.decode(BasicAuthHttpServerTest.class
-        .getResource("/auth-users.properties").getFile(), "UTF-8");
+    final String userPropertiesFile = (new File(BasicAuthHttpServerTest.class
+        .getResource("/auth-users.properties").toURI())).getPath();
```
vlsi, line 48:
```suggestion
    final String userPropertiesFile = Sources.of(BasicAuthHttpServerTest.class.getResource("/auth-users.properties"))
        .file().getAbsolutePath();
```

Please use `org.apache.calcite.util.Sources`

stoty: @vlsi 
AFAIK Avatica doesn't depend on Calcite by design, and org.apache.calcite.util.Sources is not part of Avatica.

Would you like me to copy that class into Avatica, and use it ?

vlsi: Oh, I missed it is `avatica`.
I think it makes sense to copy the class or something like that.

I do remember there is a lot of edge cases near File vs URL vs URI, so I would like to have a single class for conversion between URL and Files.

stoty: I have copied most of Sources.java and Source.java to Avatica.
I've put them in core, but it may also make sense to put them in server or even server/tests.

server/src/test/java/org/apache/calcite/avatica/server/DigestAuthHttpServerTest.java
```diff
@@ -44,8 +44,8 @@
   private static String url;
 
   @BeforeClass public static void startServer() throws Exception {
-    final String userPropertiesFile = URLDecoder.decode(BasicAuthHttpServerTest.class
-        .getResource("/auth-users.properties").getFile(), "UTF-8");
+    final String userPropertiesFile = (new File(BasicAuthHttpServerTest.class
+        .getResource("/auth-users.properties").toURI())).getPath();
```
vlsi, line 48: Please use `Sources` here as well

server/src/test/java/org/apache/calcite/avatica/server/HttpAuthBase.java
```diff
@@ -80,12 +80,11 @@ void readWriteData(String url, String tableName, Properties props) throws Except
 
   static String getHashLoginServicePropertiesString() {
     try {
-      final String userPropertiesFile =
-              URLDecoder.decode(HttpQueryStringParameterRemoteUserExtractorTest.class
-                      .getResource("/auth-users.properties").getFile(), "UTF-8");
+      final String userPropertiesFile = (new File(BasicAuthHttpServerTest.class
+          .getResource("/auth-users.properties").toURI())).getPath();
```
vlsi, line 84: `Sources` please

stoty: done

server/src/test/java/org/apache/calcite/avatica/server/DigestAuthHttpServerTest.java
stoty: done

zabetak (approved):
LGTM, if nobody objects I will merge this PR soon.

PR #148, opened 2021-06-28
[CALCITE-2322] Add fetch row count support

zacMode:
This resurrects https://github.com/apache/calcite-avatica/pull/49, updating @kminder 's existing work per guidance from [CALCITE-2322](https://issues.apache.org/jira/browse/CALCITE-2322?focusedCommentId=17370941&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17370941). 

The purpose of this change is to add support for configurable "fetch size"/ "fetch row count". Currently, setting [fetchSize](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#setFetchSize-int-) has no effect, because avatica always uses a hardcoded default value. This PR resolves that issue.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
+++ b/core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
@@ -36,6 +36,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
  */
 public abstract class AvaticaStatement
     implements Statement {
+
   /** The default value for {@link Statement#getFetchSize()}. */
   public static final int DEFAULT_FETCH_SIZE = 100;
 
@@ -108,6 +109,7 @@ public abstract class AvaticaStatement
     this.resultSetType = resultSetType;
     this.resultSetConcurrency = resultSetConcurrency;
     this.resultSetHoldability = resultSetHoldability;
+    this.fetchSize = connection.config().fetchSize(); // Default to connection config fetch size.
     this.signature = signature;
     this.closed = false;
     if (h == null) {
--- a/core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
+++ b/core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
@@ -67,7 +67,7 @@ public abstract class AvaticaStatement
   final int resultSetType;
   final int resultSetConcurrency;
   final int resultSetHoldability;
-  private int fetchSize = DEFAULT_FETCH_SIZE;
+  private int fetchRowCount;
   private int fetchDirection;
   protected long maxRowCount = 0;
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/calcite/avatica/BuiltInConnectionProperty.java
```diff
@@ -86,7 +86,10 @@
   KEY_PASSWORD("key_password", Type.STRING, "", false),
 
   HOSTNAME_VERIFICATION("hostname_verification", Type.ENUM, HostnameVerification.STRICT,
-      HostnameVerification.class, false);
+      HostnameVerification.class, false),
+
+  /** The number of rows to fetch per call, default is 100 rows. */
+  FETCH_ROW_COUNT("fetch_row_count", Type.NUMBER, AvaticaStatement.DEFAULT_FETCH_SIZE, false);
```
vlsi, line 92: What do you think of `default_fetch_size`?

zacMode: I would second @julianhyde 's  [feedback here](https://issues.apache.org/jira/browse/CALCITE-2322?focusedCommentId=17370941&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17370941):

> not "default_fetch_row_count" (because it is understood that this is a default that can be overridden per-statement; furthermore, there is a default default, which is 100);

core/src/main/java/org/apache/calcite/avatica/AvaticaStatement.java
```diff
@@ -36,6 +36,7 @@
  */
 public abstract class AvaticaStatement
     implements Statement {
+
```
vlsi, line 39: I would refrain from making unrelated changes

vlsi (approved):
LGTM modulo the blank line in AvaticaStatement

zacMode:
@F21 @zabetak @vlsi  - any update on when this can make it into a release? Thanks!

zabetak (commented):
Overall LGTM, is it possible to add some unit tests verifying the default and overriden values?

gabor-hargitai-privitar:
Hello, what is holding up this feature being merged? It would be very useful.

johndww:
Hi @zabetak / @zacMode . Any idea when this could be merged or if I could help move this through? I ran into this issue (fetch size stuck at 100) as well which isn't working for very large size results over HTTP.

core/src/main/java/org/apache/calcite/avatica/BuiltInConnectionProperty.java
```diff
@@ -86,7 +86,10 @@
   KEY_PASSWORD("key_password", Type.STRING, "", false),
 
   HOSTNAME_VERIFICATION("hostname_verification", Type.ENUM, HostnameVerification.STRICT,
-      HostnameVerification.class, false);
+      HostnameVerification.class, false),
+
+  /** The number of rows to fetch per call, default is 100 rows. */
+  FETCH_SIZE_ROWS("fetch_size_rows", Type.NUMBER, AvaticaStatement.DEFAULT_FETCH_SIZE, false);
```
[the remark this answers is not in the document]
julianhyde: Agree. The property should be called `fetch_size`. JDK has a method 'ResultSet.setFetchSize(int rows)' and so 'fetch size' is unambiguously a row count.

PR #150, opened 2021-08-24
[CALCITE-4752] PreparedStatement#SetObject() fails for BigDecimal values

stoty:
* Add BigDecimal type to lookup hash

Change-Id: I39ca2008bc3cb344fafbd3a1b8a0d21994443083

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/ColumnMetaData.java
+++ b/core/src/main/java/org/apache/calcite/avatica/ColumnMetaData.java
@@ -26,6 +26,7 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.google.protobuf.Descriptors.FieldDescriptor;
 
 import java.lang.reflect.Type;
+import java.math.BigDecimal;
 import java.sql.Array;
 import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
@@ -341,7 +342,7 @@ public class ColumnMetaData {
 
     /** Values are represented as some sub-class of {@link Number}.
      * The JSON encoding does this. */
-    NUMBER(Number.class, Types.NUMERIC),
+    NUMBER(BigDecimal.class, Types.NUMERIC),
 
     ARRAY(Array.class, Types.ARRAY),
     MULTISET(List.class, Types.JAVA_OBJECT),
--- a/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
@@ -118,7 +118,7 @@ public abstract class AbstractCursor implements Cursor {
     case Types.DOUBLE:
       return new DoubleAccessor(getter);
     case Types.DECIMAL:
-      return new BigDecimalAccessor(getter);
+      return new NumberAccessor(getter, columnMetaData.scale);
     case Types.CHAR:
       switch (columnMetaData.type.rep) {
       case PRIMITIVE_CHAR:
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/apache/calcite/avatica/ColumnMetaData.java
```diff
@@ -360,6 +361,8 @@ public ColumnMetaData setRep(Rep rep) {
         builder.put(rep.clazz, rep);
       }
       builder.put(byte[].class, BYTE_STRING);
+      //The lookup code is not smart enough to handle subclasses
+      builder.put(BigDecimal.class, NUMBER);
```
joshelser, line 365: Any reason to not put this as a value for the `Rep` enum?

server/src/test/java/org/apache/calcite/avatica/remote/RemoteMetaTest.java
```diff
@@ -481,6 +483,38 @@ private static String longString(String fragment, int length) {
...
+
+        pstmt.setString(1, "3");
+        pstmt.setObject(2, new BigDecimal("12345.67890"), Types.NUMERIC);
+        assertEquals(1, pstmt.executeUpdate());
+
+        pstmt.setString(1, "4");
+        pstmt.setObject(2, new BigDecimal("12345.67890"), Types.DECIMAL);
+        assertEquals(1, pstmt.executeUpdate());
+      }
+      conn.commit();
```
joshelser, line 514: Wouldn't be a bad idea to validate that we can pull the value back out as a BigDecimal too (if I messed up BigDecimal on the write path, wouldn't be surprised if I also did on the read path :P).

stoty: Sure, I'll add that tomorrow.

core/src/main/java/org/apache/calcite/avatica/ColumnMetaData.java
stoty: Adding   `BIG_DECIMAL(BigDecimal.class, Types.NUMERIC),` works.

However, ColumnMetaData.Rep seems to be mapped to org.apache.calcite.avatica.proto.Common.Rep, so that would 
change the wire representation, which may cause compatibility issues.

I admit that I haven't traced through the whole stack, but there are multiple comments that we want to map everything to NUMBER, so it may not be desirable.

joshelser: Yeah, this is a rats nest that I am mostly responsible for. The type-wrangling is the worst part about Avatica (especially with trying to support both json and protobuf). The Javadoc on TypedValue is the best explanation I can give as to the different "representations" of data on the wire that Avatica is following.

server/src/test/java/org/apache/calcite/avatica/remote/RemoteMetaTest.java
stoty: Added the tests

core/src/main/java/org/apache/calcite/avatica/ColumnMetaData.java
stoty: So I spent some more time looking at this.

AFAICT, adding a new ColumnMetaData.Rep value would not affect backwards compatibility with Protobuf, as we're already putting and expecting Common.Rep.BIG_DECIMAL on the wire anyway.

However,  with JSON we do put (the string value of) ColumnMetaData.Rep on the wire and try to parse it on the other side, so adding a new a value (using BIG_DECIMAL instead of NUMBER) would break backwards compatibility with older JSON servers, which could not parse that.

see https://calcite.apache.org/avatica/docs/json_reference.html#typedvalue

The current solution doesn't break backwards comatibility

WDYT @joshelser ?

core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -722,13 +701,21 @@ protected Number getNumber() throws SQLException {
       return (Number) super.getObject();
     }
 
+    //FIXME There are several problems with this, the code below simply implements
+    //a previous behaviour codified by the Calcite test suite.
+    //
+    // 1. It interprets a scale of 0 as a NOOP parameter, it should in fact drop all fractionals
```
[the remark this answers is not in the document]
joshelser: Yeah, that is funny. Julian might have opinions if you have an interest in knowing why it is this way. With a lot of this type handling, it's been better (IMO) to make a decision based on what we know at the current time :)

PR #152, opened 2021-08-28
[CALCITE-4757] In Avatica, support columns of type "NULL" in query results(NobiGo)

NobiGo:


zinking:
does avatica side require change ? or simply waiting for calcite version bump?

NobiGo:
> does avatica side require change ? or simply waiting for calcite version bump?

This issue needs to merge avatica PR first. Then upgrade the avatica version in calcite.

PR #161, opened 2021-11-06
[CALCITE-4877] Make the exception information more explicit for instantiate plugin.

sunjincheng121:
I want make the exception information more explicit for instantiate plugin in this PR. such as when we want REATE SCHEMA or CREATE TABLE supported. and we do not add the class to the class path we got the exception as follows:

`Error: Error while executing SQL "CREATE TABLE t (i INTEGER, j VARCHAR(10))": Property 'org.apache.calcite.sql.parser.ddl.SqlDdlParserImpl#FACTORY' not valid for plugin type org.apache.calcite.sql.parser.SqlParserImplFactory (state=,code=0)`

 would be great if we can make the error message more explicit , such as: 

`"Property 'org.apache.calcite.sql.parser.ddl.SqlDdlParserImpl#FACTORY' not valid as 'org.apache.calcite.sql.parser.ddl.SqlDdlParserImpl' not found in the classpath."  `

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
+++ b/core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
@@ -209,26 +209,23 @@ public class AvaticaUtils {
    */
   public static <T> T instantiatePlugin(Class<T> pluginClass,
       String className) {
+    String right = null;
+    String left = null;
     try {
       // Given a static field, say "com.example.MyClass#FOO_INSTANCE", return
       // the value of that static field.
       if (className.contains("#")) {
-        try {
-          int i = className.indexOf('#');
-          String left = className.substring(0, i);
-          String right = className.substring(i + 1);
-          //noinspection unchecked
-          final Class<T> clazz = (Class) Class.forName(left);
-          final Field field;
-          field = clazz.getField(right);
-          return pluginClass.cast(field.get(null));
-        } catch (NoSuchFieldException e) {
-          // ignore
-        }
+        int i = className.indexOf('#');
+        left = className.substring(0, i);
+        right = className.substring(i + 1);
+        //noinspection unchecked
+        final Class<T> clazz = (Class) Class.forName(left);
+        final Field field;
+        field = clazz.getField(right);
+        return pluginClass.cast(field.get(null));
       }
       //noinspection unchecked
       final Class<T> clazz = (Class) Class.forName(className);
-      assert pluginClass.isAssignableFrom(clazz);
       try {
         // We assume that if there is an INSTANCE field it is static and
         // has the right type.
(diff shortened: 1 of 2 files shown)


sunjincheng121:
`Error:   1.3sec org.apache.calcite.test.ServerQuidemTest > test(String)[5], [5] sql/schema.iq` will be fixed later.

core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
```diff
@@ -235,10 +232,29 @@ public static Class box(Class clazz) {
       } catch (NoSuchFieldException e) {
         // ignore
       }
+      assert pluginClass.isAssignableFrom(clazz);
```
vlsi, line 235: Assertions are not enabled by default. Would you please replace the assertion with a regular if check?

sunjincheng121: +1 for your suggestion. 
I do not want add assert here, just moved the position from up line to here . Agree with you, I do not like use assert to do the logical judgment.  but I see that the project seems to favor to use assert. For example, in the following classes: `avaticaresultset`, `avaticasite`, `avaticaconnection`, `builtinconnectionproperty`, `connectionconfigimpl` etc. 
Do you know any principles for using assert in this project? // I'm new to this project :)

sunjincheng121:
Thanks for you review @vlsi !
I have updated PR accordingly, I appreciate if you can have another look, Thank you!
Best，
Jincheng

core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
asolimando: My rule of thumb is to use assertions for situations that should never happen, it's more of a "documentation" for me where I state expectations that follow logically but are maybe not evident, all the other cases need checks+exception

vlsi:
Thanks. I think the changes are good, however would you please adjust test data to fix test failures?

```
FAILURE   1.1sec, org.apache.calcite.test.ServerQuidemTest > test(String)[5], [5] sql/schema.iq
    org.opentest4j.AssertionFailedError: Files differ: /home/runner/work/calcite-avatica/calcite/server/build/resources/test/sql/surefire/sql/schema.iq /home/runner/work/calcite-avatica/calcite/server/build/resources/test/sql/schema.iq
    69c69,229
    < Property 'com.example.BadSchemaFactory' not valid for plugin type org.apache.calcite.schema.SchemaFactory
    ---
    > java.sql.SQLException: Error while executing SQL "create foreign schema fs library 'com.example.BadSchemaFactory'": Property 'com.example.BadSchemaFactory' not valid as 'com.example.BadSchemaFactory' not found in the classpath
```

vlsi (changes requested):
Please fix test failures. The rest looks good.

sunjincheng121:
> Thanks. I think the changes are good, however would you please adjust test data to fix test failures?
```
FAILURE   1.1sec, org.apache.calcite.test.ServerQuidemTest > test(String)[5], [5] sql/schema.iq
    org.opentest4j.AssertionFailedError: Files differ: /home/runner/work/calcite-avatica/calcite/server/build/resources/test/sql/surefire/sql/schema.iq /home/runner/work/calcite-avatica/calcite/server/build/resources/test/sql/schema.iq
    69c69,229
    < Property 'com.example.BadSchemaFactory' not valid for plugin type org.apache.calcite.schema.SchemaFactory
    ---
    > java.sql.SQLException: Error while executing SQL "create foreign schema fs library 'com.example.BadSchemaFactory'": Property 'com.example.BadSchemaFactory' not valid as 'com.example.BadSchemaFactory' not found in the classpath
```
Thanks for your feedback @vlsi 
I think the test failure is caused by the interdependence between the calcite project and the current calcite-avatica project. After the the PR （ https://github.com/apache/calcite/pull/2600 ) merged, the error will be solved automatically? If there is any thing with my understanding, please correct me. :)

core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
sunjincheng121: I see. Thanks for sharing your thoughts on the use of assert @asolimando , Thank you!

PR #162, opened 2021-11-15
[CALCITE-4121] Driver misplaces properties from URL while connecting (Laksh Singla)

### LINKED ISSUE #10184: JDBC: Support context parameters sent through connection string
In JDBC there are two common ways of setting connection properties, which are used to pass query context parameters through to Druid SQL and Druid native queries. (Additionally, some of them, like "user" and "password", are used for internal purposes by the driver.)

1. Properties object "props" passed into `DriverManager.getConnection(url, props)`.
2. Semicolon-separated key=value pairs appended to the "url" parameter of the same method.

Today (0.18.0) Druid supports (1) but not (2). It would be good to support (2) as well, since while some third-party JDBC applications support both methods of specifying properties, some only support (2).

This is something Avatica should be handling for us, but there is a bug that prevents it from working: https://issues.apache.org/jira/browse/CALCITE-4121.

LakshSingla:
`Driver.java` fails to pass the parameters passed via connection URL to the `OpenConnectionRequest`. This PR introduces changes which extract the context properties, and pass it to the `OpenConnectionRequest` call. 

Fixes: apache/druid#10184
Resolves: https://issues.apache.org/jira/browse/CALCITE-4121

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/remote/Driver.java
+++ b/core/src/main/java/org/apache/calcite/avatica/remote/Driver.java
@@ -18,6 +18,7 @@ package org.apache.calcite.avatica.remote;
 
 import org.apache.calcite.avatica.AvaticaConnection;
 import org.apache.calcite.avatica.BuiltInConnectionProperty;
+import org.apache.calcite.avatica.ConnectStringParser;
 import org.apache.calcite.avatica.ConnectionConfig;
 import org.apache.calcite.avatica.ConnectionProperty;
 import org.apache.calcite.avatica.DriverVersion;
@@ -173,9 +174,13 @@ public class Driver extends UnregisteredDriver {
     // super.connect(...) should be creating a service and setting it in the AvaticaConnection
     assert null != service;
 
+    // extract the parameters passed in the URL info
+    final String urlSuffix = url.substring(getConnectStringPrefix().length());
+    final Properties info2 = ConnectStringParser.parse(urlSuffix, info);
+
     service.apply(
         new Service.OpenConnectionRequest(conn.id,
-            Service.OpenConnectionRequest.serializeProperties(info)));
+            Service.OpenConnectionRequest.serializeProperties(info2)));
 
     return conn;
   }


joshelser:
Thanks for trying to fix this, @LakshSingla 

I'm not 100% certain anymore, but this may have been an intentional decision previously. Avatica has its own set of JDBC properties which can be passed into the driver which control how the Avatica JDBC driver can talk to the Avatica server. These properties can be specified either in Java via the `Properties` object in `DriverManager` or via the JDBC url itself as key-value pairs.

https://calcite.apache.org/avatica/docs/client_reference.html

In addition to that, there are _also_ JDBC properties which the Avatica server (with the real backend JDBC driver) may want/need to talk to the backend database. I believe your hope is to make it such that Avatica clients could directly pass JDBC properties to the backend database.

I remember a worry I had: "what if a property name we use for Avatica conflicts with a property for the backend database?". I could see past-Josh having made the decision to not directly pass properties from client, through Avatica server, to the backend database.

Would it be possible for Avatica to only pass Properties which it does not know about into the backend database rather than all Properties?

We definitely need some test-case coverage here. We use HSQLDB for all of the unit tests against a "real" database, but I do not see any option which you could provide to do an end-to-end test case http://hsqldb.org/doc/2.0/guide/dbproperties-chapt.html#N16054

I would suggest you start looking in https://github.com/apache/calcite-avatica/blob/master/server/src/test/java/org/apache/calcite/avatica/RemoteDriverTest.java to see if there is a natural place you can validate the server-side Connection Properties. https://github.com/apache/calcite-avatica/blob/master/server/src/test/java/org/apache/calcite/avatica/remote/ConnectionPropertiesTest.java may also be helpful. There are not too many test classes in the server module -- that is where you should look.

LakshSingla:
Thanks for the insightful comment @joshelser. Before I try to make any code changes, I wanted to check if my understanding is correct.

There are 2 types of properties:
1. Those which define how the Avatica driver communicates with the database. 
2. Custom properties from the Avatica client to the server.

According to your comment, it seems that the first set of properties could be passed in a) JDBC URL or b). While getting the connection from DriverManager as `DriverManager.getConnection(url, PROPERTIES)`, but there is no provision for setting the second set of __custom__ properties from client to the DB. As per my understanding however, the second set of __custom__ properties can be passed in the properties object, but not via the URL. 
Is this behaviour intentional or accidental?

PR #165, opened 2021-12-11
CALCITE-4933 Prepare for 1.20

joshelser:
Walked through the steps at https://calcite.apache.org/avatica/docs/howto.html#making-a-release-candidate-for-calcite-committers

I think this should be everything?

### DIFF
--- a/README
+++ b/README
@@ -1,4 +1,4 @@
-Apache Calcite Avatica release 1.19.0
+Apache Calcite Avatica release 1.20.0
 
 # Overview
 This is a source or binary distribution of Avatica, a framework for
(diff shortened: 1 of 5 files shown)


julianhyde:
Sorry, I never even saw this PR yesterday. Maybe because you used "CALCITE-4933" rather than "[CALCITE-4933]" the bot didn't link it to the JIRA case.

The volume of changes in GitHub is so high that I have notifications from GitHub switched off. If it's not a JIRA change or an email to dev@ I don't see it.

joshelser:
@julianhyde sorry about that. I'm used to the PR-linker breaking when the brackets are included. Not a problem, sorry I didn't save you any cycles.

PR #166, opened 2021-12-13
Silence standard out messages in tests

zabetak:
Before this change running the tests prints a lot of messages to
standard out cluttering useful output (testname, success, failure, etc.)
and slowing down the build.

1. Remove direct calls to System.out in tests; it is considered bad
practice in general.
2. Remove SPNEGO debug information by unsetting System properties; when
necessary the developer can set them explicitly. Debug info shouldn't
be always on especially on standard out.
3. Use loggers instead of System.out to print useful info in production
code.


asolimando (approved):
LGTM

NobiGo (approved):
LGTM

zabetak:
Thanks for the reviews @asolimando @NobiGo !

PR #170, opened 2021-12-30
[CALCITE-4970] Support type conversion when no conversion necessary

knut-forkalsrud:


core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -423,6 +423,10 @@ public Reader getNCharacterStream() throws SQLException {
     }
 
     public <T> T getObject(Class<T> type) throws SQLException {
+      Object val = getObject();
+      if (val == null || type == null || type.isAssignableFrom(val.getClass())) {
```
knut-forkalsrud, line 427: or possibly `type.isInstance(val)`

PR #173, opened 2022-03-29
CALCITE-5009 Transparent JDBC connection re-creation may lead to data…

stoty:
… loss

Change-Id: Iaf10b6e4d4760cce9a7fabf1762a94501684f23d


stoty:
I don't think that the checkstyle/rat errors are related to this patch.

F21:
I've fixed the build, you should be able to rebase on top of master to get the tests passing.

stoty:
thank you @F21.

joshelser (approved):
I was going to suggest that we also update https://github.com/apache/calcite-avatica/blob/master/site/_docs/client_reference.md but I see that it's already out of date. We can merge this change and then go back and make sure the client_reference.md page is updated for all of the user-facing configuration properties (maybe adding the client retries property there, too).

PR #177, opened 2022-04-19
[CALCITE-5102] Update github-pages gem for site build

michaelmior:
It looks as though the Docker image being used for Jekyll was updated to Ruby 3 and the version of the `github-pages` gem we were using is too old. If someone else could test and see if this works, that would be appreciated.


F21:
LGTM! Tested locally and the site builds again 🎉

PR #178, opened 2022-04-20
[CALCITE-5095] Support Java 18 and Guava 31.1-jre

libenchao:


.github/workflows/main.yml
```diff
@@ -78,6 +78,21 @@ jobs:
       run: |
         ./gradlew --no-parallel --no-daemon build javadoc
 
+  linux:
```
zabetak, line 81: The GitHub actions are not running in this PR and it seems that the problem is caused by this line. I think you need to pick a unique label.

Moreover, I am not sure if you need to define a new job or just change the one above to use `java-version: 18`.

zabetak (changes requested):

.github/workflows/main.yml
libenchao: @zabetak thanks for your review. I added this job because JDK 8/11/17 are LTS versions, and we better test against all LTS versions and latest version.  
However, I found below 'mac' job's java version is 14, maybe we can just upgrade it to 18, WHYT?

zabetak (approved):
LGTM, I will merge this now!

PR #183, opened 2022-09-22
[CALCITE-5295] Read the values of plugins (such as connect string properties) from ThreadLocal fields

julianhyde:


core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
```diff
@@ -222,15 +223,22 @@ public static <T> T instantiatePlugin(Class<T> pluginClass,
...
       }
       //noinspection unchecked
       final Class<T> clazz = (Class) Class.forName(className);
       try {
         // We assume that if there is an INSTANCE field it is static and
         // has the right type.
         final Field field = clazz.getField("INSTANCE");
-        return pluginClass.cast(field.get(null));
+        value = field.get(null);
+        return pluginClass.cast(value);
```
libenchao, line 241: In this branch, `value` could also be a `ThreadLocal`?

core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
```diff
@@ -211,6 +211,7 @@ public static <T> T instantiatePlugin(Class<T> pluginClass,
       String className) {
     String right = null;
     String left = null;
+    Object value = null;
```
libenchao, line 214: Does`AvaticaUtils#instantiatePlugin` and `ConnectionConfigImpl#getPlugin`'s java doc need to explicitly say that the 'INSTANCE' could be a `ThreadLocal`?

libenchao (commented):
@julianhyde Thanks for your PR, generally looks good, I've left two minor comments.

core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
julianhyde: No. If a class `Foo` has a field called `INSTANCE`, it should be of type `Foo`. It would be confusing for a field called `INSTANCE` to be of any other type.

julianhyde: Good question, though.

core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
julianhyde: In fa3244a99 I have improved the javadoc for the 'Class#FOO' case.

I don't think we should allow `INSTANCE` to be a `ThreadLocal` because in the singleton pattern INSTANCE should have the same type as its enclosing class.

core/src/main/java/org/apache/calcite/avatica/AvaticaUtils.java
libenchao: Agreed, thanks for the explanation.

libenchao (approved):
@julianhyde Thanks for addressing my comment, now the PR LGTM!

PR #184, opened 2022-10-17
CALCITE-5327 Make SSL key-store type configurable

richardantal:


server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
```diff
@@ -850,6 +875,9 @@ protected SslContextFactory.Server buildSSLContextFactory() {
         sslFactory.setKeyStorePassword(keystorePassword);
         sslFactory.setTrustStorePath(truststore.getAbsolutePath());
         sslFactory.setTrustStorePassword(truststorePassword);
+        if (keystoreType != null && !keystoreType.equals(DEFAULT_KEYSTORE_TYPE)) {
+          sslFactory.setKeyStoreType(keystoreType);
```
joshelser, line 879: Could you add a comment here as to why this is important? Was there some use-case which can accept a keystore in a different type?

F21:
+1 to @joshelser's comments, otherwise everything else looks great!

server/src/main/java/org/apache/calcite/avatica/server/HttpServer.java
richardantal: Jetty supports other key-store formats as well, for example BCFKS key format. BCFKS is FIPS compatible but the default JKS is not.
This way we will be able configure the key-store format from phoenix-queryserver.

joshelser: thanks for the explanation.

joshelser:
Please update the website with the new user-facing API and then I think this is 👍🏼

F21:
Also, please squash your commits as well 😄

F21 (approved):
LGTM! Thanks @richardantal and @joshelser !

richardantal:
Thank you for the reviews!

joshelser:
Thanks Richie! Appreciate you merging, Francis!

PR #186, opened 2022-10-24
[CALCITE-5308] Add support for TIMESTAMP WITH LOCAL TIME ZONE and microsecond-precision

wnob:
Have the cursor interpret a `BigDecimal` in a time/date field as *seconds since epoch*, with up to nanosecond precision. Also, implement a proper getter for `TIMESTAMP WITH LOCAL TIME ZONE` which is given proper Calcite support in https://github.com/apache/calcite/pull/2973.


wnob:
Superceded by https://github.com/apache/calcite-avatica/pull/205 which focuses just on the type support and punts on precision for now.

PR #197, opened 2022-11-15
[CALCITE-2989] Use ISO calendar system when accessing Date/Time/Timestamp from a number

freastro:
This changes the unix timestamp to Date/Time/Timestamp conversion in `AbstractCursor` to convert between the ISO calendar system and the standard Gregorian calendar. This ensures that unix timestamps produced by `DateTimeUtils` are correctly converted to `java.sql.Date/Time/Timestamp` objects.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
@@ -145,7 +145,7 @@ public abstract class AbstractCursor implements Cursor {
       case NUMBER:
         return new DateFromNumberAccessor(getter, localCalendar);
       case JAVA_SQL_DATE:
-        return new DateAccessor(getter);
+        return new DateAccessor(getter, localCalendar);
       default:
         throw new AssertionError("bad " + columnMetaData.type.rep);
       }
@@ -156,7 +156,7 @@ public abstract class AbstractCursor implements Cursor {
       case NUMBER:
         return new TimeFromNumberAccessor(getter, localCalendar);
       case JAVA_SQL_TIME:
-        return new TimeAccessor(getter);
+        return new TimeAccessor(getter, localCalendar);
       default:
         throw new AssertionError("bad " + columnMetaData.type.rep);
       }
@@ -167,7 +167,7 @@ public abstract class AbstractCursor implements Cursor {
       case NUMBER:
         return new TimestampFromNumberAccessor(getter, localCalendar);
       case JAVA_SQL_TIMESTAMP:
-        return new TimestampAccessor(getter);
+        return new TimestampAccessor(getter, localCalendar);
       case JAVA_UTIL_DATE:
         return new TimestampFromUtilDateAccessor(getter, localCalendar);
       default:
@@ -260,27 +260,6 @@ public abstract class AbstractCursor implements Cursor {
       v -= calendar.getTimeZone().getOffset(v);
     }
     return DateTimeUtils.unixTimeToString(v);
-  }
-
-  private static Date longToDate(long v, Calendar calendar) {
-    if (calendar != null) {
-      v -= calendar.getTimeZone().getOffset(v);
-    }
-    return new Date(v);
-  }
-
-  static Time intToTime(int v, Calendar calendar) {
-    if (calendar != null) {
-      v -= calendar.getTimeZone().getOffset(v);
-    }
-    return new Time(v);
-  }
-
-  static Timestamp longToTimestamp(long v, Calendar calendar) {
-    if (calendar != null) {
-      v -= calendar.getTimeZone().getOffset(v);
-    }
-    return new Timestamp(v);
   }
 
   /** Implementation of {@link Cursor.Accessor}. */
@@ -908,11 +887,15 @@ public abstract class AbstractCursor implements Cursor {
    * Accessor that assumes that the underlying value is a DATE,
    * in its default representation {@code int};
    * corresponds to {@link java.sql.Types#DATE}.
+   *
+   * <p>For backwards compatibility, timezone offsets are calculated in relation to the local
+   * timezone instead of to UTC. Providing the default calendar will return the underlying value
+   * unmodified. This differs from {@link DateAccessor} which uses the timezone offset from UTC.
    */
-  private static class DateFromNumberAccessor extends NumberAccessor {
+  static class DateFromNumberAccessor extends NumberAccessor {
     private final Calendar localCalendar;
 
-    private DateFromNumberAccessor(Getter getter, Calendar localCalendar) {
+    DateFromNumberAccessor(Getter getter, Calendar localCalendar) {
       super(getter, 0);
       this.localCalendar = localCalendar;
     }
@@ -926,7 +909,7 @@ public abstract class AbstractCursor implements Cursor {
       if (v == null) {
         return null;
       }
-      return longToDate(v.longValue() * DateTimeUtils.MILLIS_PER_DAY, calendar);
+      return DateTimeUtils.unixDateToSqlDate(v.intValue(), calendar);
     }
 
     @Override public Timestamp getTimestamp(Calendar calendar) throws SQLException {
@@ -934,7 +917,8 @@ public abstract class AbstractCursor implements Cursor {
       if (v == null) {
         return null;
       }
-      return longToTimestamp(v.longValue() * DateTimeUtils.MILLIS_PER_DAY,
+      return DateTimeUtils.unixTimestampToSqlTimestamp(
+          v.longValue() * DateTimeUtils.MILLIS_PER_DAY,
           calendar);
     }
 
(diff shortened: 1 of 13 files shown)


asolimando:
@freastro, tests should pass in all TZs, you cannot assume a specific one so you can't hardcode your expected values, you need to derive them dynamically, see if [CALCITE-4793](https://issues.apache.org/jira/browse/CALCITE-4793) can help

core/src/test/java/org/apache/calcite/avatica/util/TimestampFromUtilDateAccessorTest.java
```diff
@@ -0,0 +1,221 @@
...
+import java.util.Locale;
+import java.util.SimpleTimeZone;
+import java.util.TimeZone;
+
+import static org.apache.calcite.avatica.util.DateTimeUtils.MILLIS_PER_HOUR;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class TimestampFromUtilDateAccessorTest {
```
julianhyde, line 36: Please ensure that every class (including test classes and inner classes) has a javadoc comment. It will help future maintainers understand the purpose of the class.

core/src/test/java/org/apache/calcite/avatica/util/TimestampFromUtilDateAccessorTest.java
```diff
@@ -0,0 +1,221 @@
...
+  @Before
+  public void before() {
+    final AbstractCursor.Getter getter = new LocalGetter();
+    localCalendar = Calendar.getInstance(TimeZone.getDefault(), Locale.ROOT);
+    instance = new AbstractCursor.TimestampFromUtilDateAccessor(getter, localCalendar);
+  }
+
+  @Test
+  public void testTimestamp() throws SQLException {
+    value = new Timestamp(0L);
```
julianhyde, line 54: Please change
```
  @Test
  public void testTimestamp() throws SQLException {
```

(No 'public', and \@Test on the same line as 'void'.)
to
```
  @Test void testTimestamp() throws SQLException
```

core/src/test/java/org/apache/calcite/avatica/util/TimestampFromNumberAccessorTest.java
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  private class LocalGetter implements AbstractCursor.Getter {
+    @Override
+    public Object getObject() {
+      return value;
+    }
+
+    @Override
+    public boolean wasNull() {
```
julianhyde, line 227: \@Override should be on the same line as 'public'

core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -908,11 +887,16 @@ private byte[] getBase64Decoded() throws SQLException {
    * Accessor that assumes that the underlying value is a DATE,
    * in its default representation {@code int};
    * corresponds to {@link java.sql.Types#DATE}.
+   *
+   * <p>For backwards compatibility, timezone offsets are calculated in relation to the local
+   * timezone instead of to UTC. Providing the default calendar will return the underlying value
+   * unmodified. This differs from {@link DateAccessor} which uses the timezone offset from UTC.
+   * </p>
```
julianhyde, line 894: We generally don't use `</p>` when writing javadoc. Only `<p>`, after a blank line, as you have done already.

core/src/test/java/org/apache/calcite/avatica/util/DateAccessorTest.java
```diff
@@ -0,0 +1,137 @@
...
+  private class LocalGetter implements AbstractCursor.Getter {
+    @Override
+    public Object getObject() {
+      return value;
+    }
+
+    @Override
+    public boolean wasNull() {
+      return value == null;
+    }
```
asolimando, line 135: This will have to be changed if we finally move away from the shared `value` variable

core/src/test/java/org/apache/calcite/avatica/util/DateFromNumberAccessorTest.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.TimeZone;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class DateFromNumberAccessorTest {
+
+  private Cursor.Accessor instance;
+  private Calendar localCalendar;
+  private Object value;
```
asolimando, line 37: Same consideration as for `Date value` in the other test class

core/src/test/java/org/apache/calcite/avatica/util/TimeAccessorTest.java
```diff
@@ -0,0 +1,115 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class TimeAccessorTest {
+
+  private static final Calendar UTC =
+      Calendar.getInstance(TimeZone.getTimeZone("UTC"), Locale.ROOT);
+
+  private Cursor.Accessor instance;
+  private Calendar localCalendar;
+  private Time value;
```
asolimando, line 38: Shared variable also here, if we decide we want to play it safe, let's change this too

core/src/test/java/org/apache/calcite/avatica/util/TimeFromNumberAccessorTest.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.TimeZone;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class TimeFromNumberAccessorTest {
+
+  private Cursor.Accessor instance;
+  private Calendar localCalendar;
+  private Object value;
```
asolimando, line 37: shared r/w variable across tests

core/src/test/java/org/apache/calcite/avatica/util/TimestampAccessorTest.java
```diff
@@ -0,0 +1,216 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class TimestampAccessorTest {
+
+  private static final Calendar UTC =
+      Calendar.getInstance(TimeZone.getTimeZone("UTC"), Locale.ROOT);
+
+  private Cursor.Accessor instance;
+  private Calendar localCalendar;
+  private Timestamp value;
```
asolimando, line 42: shared r/w variable across tests

core/src/test/java/org/apache/calcite/avatica/util/TimestampFromNumberAccessorTest.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.TimeZone;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class TimestampFromNumberAccessorTest {
+
+  private Cursor.Accessor instance;
+  private Calendar localCalendar;
+  private Object value;
```
asolimando, line 38: shared r/w variable across tests

core/src/test/java/org/apache/calcite/avatica/util/TimestampFromUtilDateAccessorTest.java
```diff
@@ -0,0 +1,221 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class TimestampFromUtilDateAccessorTest {
+
+  private static final Calendar UTC =
+      Calendar.getInstance(TimeZone.getTimeZone("UTC"), Locale.ROOT);
+
+  private Cursor.Accessor instance;
+  private Calendar localCalendar;
+  private Date value;
```
asolimando, line 43: shared r/w variable across tests

core/src/test/java/org/apache/calcite/avatica/util/DateAccessorTest.java
```diff
@@ -0,0 +1,137 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class DateAccessorTest {
+
+  private static final Calendar UTC =
+      Calendar.getInstance(TimeZone.getTimeZone("UTC"), Locale.ROOT);
+
+  private Cursor.Accessor instance;
+  private Calendar localCalendar;
+  private Date value;
```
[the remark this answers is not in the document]
julianhyde: I think it's OK. If a test class has 5 test methods, then junit will create 5 instances of the class. Whether the tests are run in parallel or series, the instance variables won't conflict.

The main reason to set variables in \@Before and clear them in \@After is to prevent excessive memory leaks (and to free resources such as DB connections after the test has finished).

My personal style is to put everything on the stack of the test method, because I am suspicious of non-final fields. But I don't expect everyone to follow my style.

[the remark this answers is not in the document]
asolimando: Thanks @julianhyde for the explanation, since it's safe I leave it up to @freastro to decide what's best here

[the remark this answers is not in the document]
freastro: Thanks for the suggestion. I like to push common test code to the `@Before` method. I think it makes the tests more readable.

core/src/test/java/org/apache/calcite/avatica/util/TimestampFromUtilDateAccessorTest.java
freastro: Thanks for the suggestion. I've added a javadoc comment to every method and class.

core/src/test/java/org/apache/calcite/avatica/util/TimestampFromUtilDateAccessorTest.java
freastro: I've updated every test as requested, except for removing `public` which seems to break the tests.

core/src/test/java/org/apache/calcite/avatica/util/TimestampFromUtilDateAccessorTest.java
julianhyde: Thank you. I think that javadoc on methods is optional - especially if the methods are private/package-protected and have descriptive names, and that applies to many test methods - but the javadoc on classes is essential because once you know the purpose of the class you can often figure out the purpose of the methods.

PR #207, opened 2023-02-14
[CALCITE-5446] Add support for TIMESTAMP WITH LOCAL TIME ZONE

wnob:
I know you already suggested getting rid of the `fixedInstant` boolean. I'm going to start teasing that apart now, but I'm looking for feedback on the tests in the mean time, which make up the bulk of this PR. It's a big PR but all the test cases are relatively independent and the test files can be reviewed one at a time.

### DIFF
--- a/core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetConversionsTest.java
+++ b/core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetConversionsTest.java
@@ -50,14 +50,18 @@ import java.sql.Struct;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.sql.Types;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
-import java.util.TimeZone;
 
 import static org.hamcrest.CoreMatchers.isA;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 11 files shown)


core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetConversionsTest.java
```diff
@@ -215,15 +270,18 @@ public TestMetaImpl(AvaticaConnection connection) {
       List<Object> row = Collections.<Object>singletonList(
           new Object[] {
               true, (byte) 1, (short) 2, 3, 4L, 5.0f, 6.0d, "testvalue",
-              new Date(1476130718123L), new Time(1476130718123L),
-              new Timestamp(1476130718123L),
+              new Date(DST_INSTANT), new Time(DST_INSTANT),
```
wnob, line 273: Note:

- `DST_INSTANT = 1476130718123`
- `STANDARD_INSTANT = 1479123123242`
- `VALID_TIME = 1476123`
- `OVERFLOW_TIME = 147912242`

So this is not changing any of the test values; it's just giving them names so that equivalence is more apparent throughout the file.

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetConversionsTest.java
```diff
@@ -634,7 +701,7 @@ private DateArrayAccessorTestHelper(Getter g) {
       ColumnMetaData.ScalarType intType =
           ColumnMetaData.scalar(Types.DATE, "DATE", ColumnMetaData.Rep.PRIMITIVE_INT);
       Array expectedArray =
-          new ArrayFactoryImpl(TimeZone.getTimeZone("UTC")).createArray(
+          new ArrayFactoryImpl(DateTimeUtils.DEFAULT_ZONE).createArray(
```
wnob, line 704: This is to work around https://issues.apache.org/jira/browse/CALCITE-5488

core/src/test/java/org/apache/calcite/avatica/util/TimeAccessorTest.java
```diff
@@ -46,12 +49,13 @@ public class TimeAccessorTest {
    */
   @Before public void before() {
     final AbstractCursor.Getter getter = new LocalGetter();
-    localCalendar = Calendar.getInstance(TimeZone.getDefault(), Locale.ROOT);
-    instance = new AbstractCursor.TimeAccessor(getter, localCalendar);
+    localCalendar = Calendar.getInstance(IST_ZONE, Locale.ROOT);
```
wnob, line 52: I think it's better to use a hardcoded time zone than the system default for these unit tests, so adjustment is tested consistently regardless of the host machine.

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetConversionsTest.java
julianhyde: When I merge I might split out a 'refactor' commit that does not add or change tests but just moves literals into constant fields.

core/src/test/java/org/apache/calcite/avatica/util/TimeAccessorTest.java
julianhyde: In general I agree.

But some of JDBC's behavior is defined to use the local time zone. We need to test that behavior.

wnob: The local time zone is injected at a higher level than inside the getters. The getters are provided with the local time zone at construction time by `AbstractCursor`, so the logic involving the system default time zone is tested in `AvaticaResultSetConversionsTest`

F21:
@wob, should this be closed since there's #209?

PR #208, opened 2023-02-15
[CALCITE-5536] Clean up some of the magic numbers in AvaticaResultSetConversionsTest and AbstractCursor

wnob:
Turn a bunch of magic numbers into named constants without changing any values, and add extra comments where appropriate.

In `AvaticaResultSetConversionsTest`, I renamed `DEFAULT_CALENDAR` to `UTC_CALENDAR` because that's a more accurate name. The definition of `DateTimeUtils.calendar()` is `Calendar.getInstance(UTC_ZONE, Locale.ROOT)`.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
@@ -173,14 +173,14 @@ public abstract class AbstractCursor implements Cursor {
       default:
         throw new AssertionError("bad " + columnMetaData.type.rep);
       }
-    case 2013: // TIME_WITH_TIMEZONE
+    case Types.TIME_WITH_TIMEZONE:
       switch (columnMetaData.type.rep) {
       case STRING:
         return new StringAccessor(getter);
       default:
         throw new AssertionError("bad " + columnMetaData.type.rep);
       }
-    case 2014: // TIMESTAMP_WITH_TIMEZONE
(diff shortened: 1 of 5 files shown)


F21:
@wnob Can you open a jira issue for this?

core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -173,14 +173,14 @@ protected Accessor createAccessor(ColumnMetaData columnMetaData,
       default:
         throw new AssertionError("bad " + columnMetaData.type.rep);
       }
-    case 2013: // TIME_WITH_TIMEZONE
```
stoty, line 176: Nit:
While these changes are fine, they do not agree with the commit message. (i.e this is not a test refactor)

julianhyde: Change the commit message to "[CALCITE-5536](https://issues.apache.org/jira/browse/CALCITE-5536)
Clean up some of the magic numbers in AvaticaResultSetConversionsTest and AbstractCursor" and I think we're good.

stoty:
You have only changed the PR name, and not the commit message, @wnob.
I'm going to change the commit message, and commit this with the agreed upon commit message.

stoty:
Committed with fixed commit message.

PR #209, opened 2023-02-17
[CALCITE-5446] Add support for TIMESTAMP WITH LOCAL TIME ZONE

wnob:
Supersedes https://github.com/apache/calcite-avatica/pull/207.

Based on https://github.com/apache/calcite-avatica/pull/208, which should be merged first.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
@@ -43,6 +43,7 @@ import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.List;
 import java.util.Map;
+import java.util.TimeZone;
 
 /**
  * Base class for implementing a cursor.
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -150,26 +151,34 @@ protected Accessor createAccessor(ColumnMetaData columnMetaData,
         throw new AssertionError("bad " + columnMetaData.type.rep);
       }
     case Types.TIME:
+      // TIME WITH LOCAL TIME ZONE is a standard ISO type without proper JDBC support.
+      // It represents a global instant in time, as opposed to local clock parameters.
+      boolean fixedInstant =
```
julianhyde, line 156: 'fixedInstant' is a tautology. Call it 'instant' (which is Joda time terminology)

core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -1130,18 +1165,20 @@ static class DateAccessor extends ObjectAccessor {
    */
   static class TimeAccessor extends ObjectAccessor {
     private final Calendar localCalendar;
+    private final boolean fixedInstant;
```
julianhyde, line 1168: Remove the `fixedInstant` field and create a separate subclass. Javadoc says 'Accessor that assumes that the underlying value is a TIME' which is a lie if you make the class serve double duty for TIME and TIME_WITH_LOCAL_TIME_ZONE.

Instant and localtime are so confusingly similar we just cannot risk storing them in the same field.

Same comments apply to the TimestampAccessor; create a class TimestampLtzAccessor.

core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -1241,11 +1280,13 @@ static class TimestampAccessor extends ObjectAccessor {
    */
   static class TimestampFromUtilDateAccessor extends ObjectAccessor {
     private final Calendar localCalendar;
+    private final boolean fixedInstant;
 
-    TimestampFromUtilDateAccessor(Getter getter,
-        Calendar localCalendar) {
+    TimestampFromUtilDateAccessor(
+        Getter getter, Calendar localCalendar, boolean fixedInstant) {
       super(getter);
```
julianhyde, line 1287: This seems impossible. How could a `java.util.Date` represent anything but an instant?

core/src/test/java/org/apache/calcite/avatica/util/TimeAccessorTest.java
```diff
@@ -46,77 +49,74 @@ public class TimeAccessorTest {
...
-    localCalendar = Calendar.getInstance(TimeZone.getDefault(), Locale.ROOT);
-    instance = new AbstractCursor.TimeAccessor(getter, localCalendar);
+    localCalendar = Calendar.getInstance(IST_ZONE, Locale.ROOT);
+    instance = new AbstractCursor.TimeAccessor(getter, localCalendar, false);
   }
 
   /**
-   * Test {@code getTime()} returns the same value as the input time for the local calendar.
+   * Test {@code getTime()} returns the same value as the input time for the connection default
+   * calendar.
```
julianhyde, line 58: Is there a bug in Avatica? If so, say what it is. If not, why are you changing/removing tests?

core/src/test/java/org/apache/calcite/avatica/util/TimeAccessorTest.java
```diff
@@ -46,77 +49,74 @@ public class TimeAccessorTest {
...
-   * UTC.
-   */
+  /** Test {@code getTime()} handles time zone conversions relative to the provided calendar. */
   @Test public void testTimeWithCalendar() throws SQLException {
     value = new Time(0L);
 
     final TimeZone minusFiveZone = TimeZone.getTimeZone("GMT-5:00");
     final Calendar minusFiveCal = Calendar.getInstance(minusFiveZone, Locale.ROOT);
-    assertThat(instance.getTime(minusFiveCal).getTime(),
+    assertThat(
```
julianhyde, line 71: reformatting isn't helpful

core/src/test/java/org/apache/calcite/avatica/util/TimeFromNumberAccessorTest.java
```diff
@@ -36,112 +35,90 @@
...
   }
 
   /**
-   * Test {@code getTime()} handles time zone conversions relative to the local calendar and not
-   * UTC.
+   * Test {@code getTime()} and {@code getTimestamp()} handle time zone conversions relative to the
+   * provided calendar.
    */
   @Test public void testTimeWithCalendar() throws SQLException {
-    final int offset = localCalendar.getTimeZone().getOffset(0);
```
julianhyde, line 81: so many test changes without any justification. I'm giving up.

core/src/test/java/org/apache/calcite/avatica/util/TimeAccessorTest.java
wnob: This was just me trying to reconcile the test description with the logic therein, though I realize now that I didn't do a great job of it. The test says "for the local calendar", but it's unclear if that means `localCalendar` (which is not involved in this test at all), or UTC (which in most cases would not be accurately described as "the local calendar"). Upon closer inspection I can appreciate that this test case is really meant to test that no adjustment occurs when the explicitly provided calendar is either UTC or null. I'm working to clean up these tests in general, and in this case I think the I should've just updated the javadoc.

PR #213, opened 2023-03-14
[CALCITE-5567] Update mockito from 4.4.0 to 4.11.0 and enable jdk19

snuyanzin:
The PR makes it build with jdk19 instead of jdk18
It also bumps mockito since mockito 4.4.0 depends on byte buddy which is not compatible with jdk19


asolimando (approved):
LGTM

snuyanzin:
Thanks for having a look @asolimando

PR #217, opened 2023-03-22
[CALCITE-5581] Implement Basic client side load balancing in Avatica Driver

vaijosh:
CALCITE-5581
Added a basic client side load balancing behaviour which allows user specify list of URLs and get client connection to any of the URL using Load balancing strategy. Out of the box we provided RandomSelect and RoundRobin LB strategies, but user can implement their own and plug it using LB_STRATEGY parameter. Tested the changes with actual phoenix query thin client driver.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/BuiltInConnectionProperty.java
+++ b/core/src/main/java/org/apache/calcite/avatica/BuiltInConnectionProperty.java
@@ -16,8 +16,11 @@
  */
 package org.apache.calcite.avatica;
 
+import org.apache.calcite.avatica.ha.ShuffledRoundRobinLBStrategy;
 import org.apache.calcite.avatica.remote.AvaticaHttpClientFactoryImpl;
 import org.apache.calcite.avatica.remote.HostnameVerificationConfigurable.HostnameVerification;
+
+import org.apache.hc.core5.util.Timeout;
 
 import java.util.HashMap;
 import java.util.HashSet;
@@ -33,6 +36,7 @@ import static org.apache.calcite.avatica.ConnectionConfigImpl.parse;
  * Enumeration of Avatica's built-in connection properties.
  */
 public enum BuiltInConnectionProperty implements ConnectionProperty {
+
   /** Factory. */
   FACTORY("factory", Type.PLUGIN, null, false),
 
@@ -91,7 +95,36 @@ public enum BuiltInConnectionProperty implements ConnectionProperty {
   TRANSPARENT_RECONNECTION("transparent_reconnection", Type.BOOLEAN, Boolean.FALSE, false),
 
   /** Number of rows to fetch per call. */
-  FETCH_SIZE("fetch_size", Type.NUMBER, AvaticaStatement.DEFAULT_FETCH_SIZE, false);
+  FETCH_SIZE("fetch_size", Type.NUMBER, AvaticaStatement.DEFAULT_FETCH_SIZE, false),
+
+  /** Avatica connection HA property  - use client side load balancing **/
+  USE_CLIENT_SIDE_LB("use_client_side_lb", Type.BOOLEAN, Boolean.FALSE, false),
+
+  /** Avatica connection HA property  - Load balanced URLs **/
+  LB_URLS("lb_urls", Type.STRING, "", false),
+
+  /** Avatica connection HA property  - Load balancing strategy **/
+  LB_STRATEGY("lb_strategy", Type.PLUGIN,
+      ShuffledRoundRobinLBStrategy.class.getName(), false),
+
+  /**
+   * The number of retries we need for failover during client side load balancing.
+   */
+  LB_CONNECTION_FAILOVER_RETRIES("lb_connection_failover_retries",
+      Type.NUMBER, 3, false),
+
+  /**
+   * The amount of time in millis that the driver should wait before attempting
+   * connection failover
+   */
+  LB_CONNECTION_FAILOVER_SLEEP_TIME("lb_connection_failover_sleep_time",
+      Type.NUMBER, 1000, false),
+
+  /**
+   * HTTP Connection Timeout in milliseconds.
+   */
+  HTTP_CONNECTION_TIMEOUT("http_connection_timeout",
+      Type.NUMBER, Timeout.ofMinutes(3).toMilliseconds(), false);
 
   private final String camelName;
   private final Type type;
@@ -117,6 +150,7 @@ public enum BuiltInConnectionProperty implements ConnectionProperty {
     for (BuiltInConnectionProperty p : BuiltInConnectionProperty.values()) {
       LOCAL_PROPS.add(p.camelName());
     }
+
   }
 
   BuiltInConnectionProperty(String camelName, Type type, Object defaultValue,
--- a/core/src/main/java/org/apache/calcite/avatica/ConnectionConfig.java
+++ b/core/src/main/java/org/apache/calcite/avatica/ConnectionConfig.java
@@ -16,6 +16,7 @@
  */
 package org.apache.calcite.avatica;
 
+import org.apache.calcite.avatica.ha.LBStrategy;
 import org.apache.calcite.avatica.remote.AvaticaHttpClientFactory;
 import org.apache.calcite.avatica.remote.HostnameVerificationConfigurable.HostnameVerification;
 import org.apache.calcite.avatica.remote.Service;
@@ -66,6 +67,18 @@ public interface ConnectionConfig {
   boolean transparentReconnectionEnabled();
   /** @see BuiltInConnectionProperty#FETCH_SIZE */
   int fetchSize();
+  /** @see BuiltInConnectionProperty#USE_CLIENT_SIDE_LB #**/
+  boolean useClientSideLb();
+  /** @see BuiltInConnectionProperty#LB_URLS **/
+  String getLbURLs();
+  /** @see BuiltInConnectionProperty#LB_STRATEGY **/
+  LBStrategy getLBStrategy();
+  /** @see BuiltInConnectionProperty#LB_CONNECTION_FAILOVER_RETRIES **/
+  int getLBConnectionFailoverRetries();
+  /** @see BuiltInConnectionProperty#LB_CONNECTION_FAILOVER_SLEEP_TIME **/
+  long getLBConnectionFailoverSleepTime();
+  /** @see BuiltInConnectionProperty#HTTP_CONNECTION_TIMEOUT **/
+  long getHttpConnectionTimeout();
 }
 
 // End ConnectionConfig.java
--- a/core/src/main/java/org/apache/calcite/avatica/ha/RandomSelectLBStrategy.java
+++ b/core/src/main/java/org/apache/calcite/avatica/ha/RandomSelectLBStrategy.java
@@ -14,25 +14,27 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.calcite.avatica.remote;
+package org.apache.calcite.avatica.ha;
 
 import org.apache.calcite.avatica.ConnectionConfig;
 
-import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Random;
 
 /**
- * Allows a http connection pool to be provided to enable TLS authentication.
- * On clients with this interface setHttpClientPool() MUST be called before using them.
+ * Random Select strategy for client side load balancing.
  */
-public interface HttpClientPoolConfigurable {
-  /**
-   * Sets a PoolingHttpClientConnectionManager containing the collection of SSL/TLS server
-   * keys and truststores to use for HTTPS calls.
-   *
-   * @param pool   The http connection pool
-   * @param config The connection config
-   */
-  void setHttpClientPool(PoolingHttpClientConnectionManager pool, ConnectionConfig config);
+public class RandomSelectLBStrategy implements LBStrategy {
+  private static final Logger LOG = LoggerFactory.getLogger(RandomSelectLBStrategy.class);
+  public static final String URL_SEPERATOR_CHAR = ",";
+  private final Random random = new Random();
+  @Override
+  public String getLbURL(ConnectionConfig config) {
+    String[] urls = config.getLbURLs().split(URL_SEPERATOR_CHAR);
+    String url = urls[random.nextInt(urls.length)];
+    LOG.info("Selected URL:{}", url);
+    return url;
+  }
 }
-
-// End HttpClientPoolConfigurable.java
--- a/core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
+++ b/core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.avatica.ha;
+
+import org.apache.calcite.avatica.ConnectionConfig;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Round Robin strategy for client side load balancing.
+ * Its implemented it as a singleton so that we can maintain state
+ * i.e. which URL was last used from the list of URLs specified.
+ */
+public class RoundRobinLBStrategy implements LBStrategy {
+  private static final Logger LOG = LoggerFactory.getLogger(RoundRobinLBStrategy.class);
+
+  public static final RoundRobinLBStrategy INSTANCE = new RoundRobinLBStrategy();
+  private RoundRobinLBStrategy() { }
+  public static final String URL_SEPERATOR_CHAR = ",";
+
+  Map<String, Integer>  configToIndexServedMap = new HashMap<>();
+  Map<String, String[]> configToUrlListMap = new HashMap<>();
+
+  @Override
+  public synchronized String getLbURL(ConnectionConfig config) {
+    String key = getKey(config);
+    if (!configToIndexServedMap.containsKey(key)) {
+      configToIndexServedMap.put(key, 0);
+      configToUrlListMap.put(key, config.getLbURLs().split(URL_SEPERATOR_CHAR));
+    }
+    String[] urls = configToUrlListMap.get(key);
+    int urlIndex = configToIndexServedMap.get(key);
+    configToIndexServedMap.put(key, (urlIndex + 1) % urls.length);
+    String url = urls[urlIndex];
+    LOG.info("Selected URL:{}", url);
+    return url;
+  }
+  private static String getKey(ConnectionConfig config) {
+    return config.getLbURLs();
+  }
+}
--- a/core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
+++ b/core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
@@ -1,0 +1,75 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.avatica.ha;
+
+import org.apache.calcite.avatica.ConnectionConfig;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+/**
+ * Shuffled Round Robin strategy for client side load balancing.
+ * It starts with some random position in a list of URLs, and then returns subsequent URL
+ * in a RoundRobin manner.
+ * It's implemented it as a singleton so that we can maintain state
+ * i.e. which URL was last used from the list of URLs specified.
+ */
+public class ShuffledRoundRobinLBStrategy implements LBStrategy {
+  private static final Logger LOG = LoggerFactory.getLogger(ShuffledRoundRobinLBStrategy.class);
+
+  public static final ShuffledRoundRobinLBStrategy INSTANCE = new ShuffledRoundRobinLBStrategy();
+  private ShuffledRoundRobinLBStrategy() { }
+  public static final String URL_SEPERATOR_CHAR = ",";
+
+  Map<String, Integer> configToIndexServedMap = new HashMap<>();
+  Map<String, String[]> configToUrlListMap = new HashMap<>();
+
+  @Override
+  public synchronized String getLbURL(ConnectionConfig config) {
+    String key = getKey(config);
+    String lbURLs = config.getLbURLs();
+    if (!configToIndexServedMap.containsKey(key)) {
+      configToIndexServedMap.put(key, 0);
+      initialiseUrlList(key, lbURLs);
+    }
+    String[] urls = configToUrlListMap.get(key);
+    int urlIndex = configToIndexServedMap.get(key);
+
+    String url = urls[urlIndex];
+    LOG.info("Selected URL:{}", url);
+    urlIndex = (urlIndex + 1) % urls.length;
+    configToIndexServedMap.put(key, urlIndex);
+    return url;
+  }
+
+  private void initialiseUrlList(String key, String lbURLs) {
+    String[] urls = lbURLs.split(URL_SEPERATOR_CHAR);
+    List<String> list = Arrays.asList(urls);
+    Collections.shuffle(list);
+    urls = list.toArray(urls);
+    configToUrlListMap.put(key, urls);
+  }
+
+  private static String getKey(ConnectionConfig config) {
+    return config.getLbURLs();
+  }
+}
--- a/core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
+++ b/core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
@@ -16,6 +16,8 @@
  */
 package org.apache.calcite.avatica.remote;
 
+import org.apache.calcite.avatica.ConnectionConfig;
+
 import org.apache.hc.client5.http.ClientProtocolException;
 import org.apache.hc.client5.http.SystemDefaultDnsResolver;
 import org.apache.hc.client5.http.auth.AuthSchemeFactory;
@@ -27,6 +29,7 @@ import org.apache.hc.client5.http.auth.KerberosCredentials;
 import org.apache.hc.client5.http.auth.StandardAuthScheme;
 import org.apache.hc.client5.http.auth.UsernamePasswordCredentials;
 import org.apache.hc.client5.http.classic.methods.HttpPost;
+import org.apache.hc.client5.http.config.RequestConfig;
 import org.apache.hc.client5.http.impl.auth.BasicAuthCache;
 import org.apache.hc.client5.http.impl.auth.BasicCredentialsProvider;
 import org.apache.hc.client5.http.impl.auth.BasicSchemeFactory;
@@ -34,6 +37,7 @@ import org.apache.hc.client5.http.impl.auth.DigestSchemeFactory;
 import org.apache.hc.client5.http.impl.auth.SPNegoSchemeFactory;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse;
+import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
 import org.apache.hc.client5.http.impl.classic.HttpClients;
 import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
 import org.apache.hc.client5.http.protocol.HttpClientContext;
@@ -57,6 +61,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.security.Principal;
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 
 /**
  * A common class to invoke HTTP requests against the Avatica server agnostic of the data being
@@ -91,11 +96,18 @@ public class AvaticaCommonsHttpClientImpl implements AvaticaHttpClient, HttpClie
     this.uri = toURI(Objects.requireNonNull(url));
   }
 
-  protected void initializeClient(PoolingHttpClientConnectionManager pool) {
+  protected void initializeClient(PoolingHttpClientConnectionManager pool,
+                                  ConnectionConfig config) {
     this.authCache = new BasicAuthCache();
     // A single thread-safe HttpClient, pooling connections via the
     // ConnectionManager
-    this.client = HttpClients.custom().setConnectionManager(pool).build();
+    RequestConfig.Builder requestConfigBuilder = RequestConfig.custom();
+    RequestConfig requestConfig = requestConfigBuilder
+        .setConnectTimeout(config.getHttpConnectionTimeout(), TimeUnit.MILLISECONDS)
+        .build();
+    HttpClientBuilder httpClientBuilder = HttpClients.custom().setConnectionManager(pool)
+        .setDefaultRequestConfig(requestConfig);
+    this.client = httpClientBuilder.build();
   }
 
   @Override public byte[] send(byte[] request) {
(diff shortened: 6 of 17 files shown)


core/src/main/java/org/apache/calcite/avatica/remote/Driver.java
```diff
@@ -119,23 +119,25 @@ Service createService(AvaticaConnection connection, ConnectionConfig config) {
...
-        break;
-      case PROTOBUF:
-        service = new RemoteProtobufService(httpClient, new ProtobufTranslationImpl());
-        break;
-      default:
-        throw new IllegalArgumentException("Unhandled serialization type: " + serializationType);
-      }
     } else {
-      service = new MockJsonService(Collections.<String, String>emptyMap());
+      if (config.url() != null) {
```
stoty, line 123: This change seems to be eqivalent to the existing code but nests one more if.
While we want to make sure that we always use curly braces in control statements, else if a special case, which actually improves readability.

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+ * Its implemented it as a singleton so that we can maintain state
+ * i.e. which URL was last used from the list of URLs specified.
+ */
+public class RoundRobinLBStrategy implements LBStrategy {
+
+  public static final RoundRobinLBStrategy INSTANCE = new RoundRobinLBStrategy();
+  private RoundRobinLBStrategy() { }
+  public static final String URL_SEPERATOR_CHAR = ",";
+
+  Map<String, Integer>  configToIndexServedMap = new HashMap<>();
```
stoty, line 36: We may consider using the config object itself, instead of just the server list.
Both have pros and cons.

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  public String getLbURL(ConnectionConfig config) {
+    try {
+      String key = getKey(config);
+      String[] urls = config.getLbURLs().split(URL_SEPERATOR_CHAR);
```
stoty, line 41: We should store the split list, and not re-process on each connection attempt.

stoty (changes requested):

stoty:
Also use square brackets in the commit message, see the other PRs for formatting.

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
vaijosh: Config was natural choice. But from callers we don't always get same object (see AvaticaConnection.config). So this was not feasible. toString() method was another option I had considered, but it was a superset of config.getLbURLs().
So config.getLbURLs() was a better option here.

core/src/main/java/org/apache/calcite/avatica/remote/Driver.java
vaijosh: Thanks @stoty.
While doing checkstyle formatting, I had accidentally done this. But restored the previous "if - else if -else" statement

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
vaijosh: Good point. I have done the changes.

vaijosh:
> Also use square brackets in the commit message, see the other PRs for formatting.

Done. Thanks.

server/src/test/java/org/apache/calcite/avatica/ha/ConnectionPropertiesHATest.java
```diff
@@ -0,0 +1,227 @@
...
+    properties.put(BuiltInConnectionProperty.LB_URLS.name(), urls);
+
+    String url = SERVERS.getJdbcUrl(START_PORT, Driver.Serialization.PROTOBUF);
+    long startTime = System.currentTimeMillis();
+    try {
+      DriverManager.getConnection(url, properties);
+    } catch (RuntimeException re) {
+      long endTime = System.currentTimeMillis();
+      long elapsedTime = endTime - startTime;
+      Assert.assertTrue(elapsedTime < Timeout.ofMinutes(3).toMicroseconds());
```
stoty, line 205: Are you comparing milli and micro seconds here ?

core/src/main/java/org/apache/calcite/avatica/ConnectionConfig.java
```diff
@@ -66,6 +67,21 @@ public interface ConnectionConfig {
   boolean transparentReconnectionEnabled();
   /** @see BuiltInConnectionProperty#FETCH_SIZE */
   int fetchSize();
+
```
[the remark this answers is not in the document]
vaijosh: Done

core/src/main/java/org/apache/calcite/avatica/ha/RandomSelectLBStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+public class RandomSelectLBStrategy implements LBStrategy {
+  public static final String URL_SEPERATOR_CHAR = ",";
+  Random random = new Random();
+  @Override
+  public String getLbURL(ConnectionConfig config) {
+    try {
+      String[] urls = config.getLbURLs().split(URL_SEPERATOR_CHAR);
+      return urls[random.nextInt(urls.length)];
+    } catch (PatternSyntaxException e) {
+      throw new RuntimeException(e);
```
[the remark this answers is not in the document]
vaijosh: We are using separator character as a pattern ( which is currently hardcoded semicolon). So we won't get PatternSyntaxException. So removed it.

core/src/main/java/org/apache/calcite/avatica/remote/Driver.java
```diff
@@ -149,8 +150,16 @@ Service createService(AvaticaConnection connection, ConnectionConfig config) {
    */
   AvaticaHttpClient getHttpClient(AvaticaConnection connection, ConnectionConfig config) {
     URL url;
+    String urlStr;
+    if (config.useClientSideLb()) {
+      urlStr = config.getLBStrategy().getLbURL(config);
+      LOG.info("LB Strategy:{}.Selected URL:{}",
```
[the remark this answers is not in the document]
vaijosh: Done.

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
```diff
@@ -91,11 +96,19 @@ public AvaticaCommonsHttpClientImpl(URL url) {
...
 
-  protected void initializeClient(PoolingHttpClientConnectionManager pool) {
+  protected void initializeClient(PoolingHttpClientConnectionManager pool,
+                                  ConnectionConfig config) {
     this.authCache = new BasicAuthCache();
     // A single thread-safe HttpClient, pooling connections via the
     // ConnectionManager
-    this.client = HttpClients.custom().setConnectionManager(pool).build();
+    RequestConfig.Builder requestConfigBuilder = RequestConfig.custom();
+    if (config.useClientSideLb()) {
```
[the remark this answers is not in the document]
vaijosh: Done.

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
```diff
@@ -91,11 +96,19 @@ public AvaticaCommonsHttpClientImpl(URL url) {
...
+      requestConfigBuilder.setConnectTimeout(config.getHttpConnectionTimeout(),
```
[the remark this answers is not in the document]
vaijosh: Done.

server/src/test/java/org/apache/calcite/avatica/ha/ConnectionPropertiesHATest.java
vaijosh: Done. Changed it to milliseconds.

core/src/main/java/org/apache/calcite/avatica/remote/Driver.java
```diff
@@ -159,25 +168,43 @@ AvaticaHttpClient getHttpClient(AvaticaConnection connection, ConnectionConfig c
...
-            Service.OpenConnectionRequest.serializeProperties(info)));
+      Service service = conn.getService();
 
-    return conn;
+      // super.connect(...) should be creating a service and setting it in the AvaticaConnection
+      assert null != service;
+      try {
+        service.apply(
+            new Service.OpenConnectionRequest(conn.id,
+                Service.OpenConnectionRequest.serializeProperties(info)));
```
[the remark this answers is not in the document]
vaijosh: Yes. I have already checked this. BuiltInConnectionProperty.isLocalProperty() decides which properties should not be serialized to the server.

stoty:
One more feature that Knox has, but this is missing is the ability to sleep before retries.
A simple flat value like Knox has should be enough.

https://docs.cloudera.com/runtime/7.2.9/knox-authentication/topics/security-knox-configure-ha.html

core/src/main/java/org/apache/calcite/avatica/BuiltInConnectionProperty.java
```diff
@@ -91,7 +95,28 @@ public enum BuiltInConnectionProperty implements ConnectionProperty {
   TRANSPARENT_RECONNECTION("transparent_reconnection", Type.BOOLEAN, Boolean.FALSE, false),
 
   /** Number of rows to fetch per call. */
-  FETCH_SIZE("fetch_size", Type.NUMBER, AvaticaStatement.DEFAULT_FETCH_SIZE, false);
+  FETCH_SIZE("fetch_size", Type.NUMBER, AvaticaStatement.DEFAULT_FETCH_SIZE, false),
+
+  /** Avatica connection HA property  - use client side load balancing **/
+  USE_CLIENT_SIDE_LB("useClientSideLB", Type.BOOLEAN, Boolean.FALSE, false),
```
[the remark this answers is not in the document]
vaijosh: After changing the property names to under score format the properties are not reaching to the retry logic block ( which is dependent of serialized properties). So I need to investigate this.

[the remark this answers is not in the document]
vaijosh: Fixed the issue.

vaijosh:
> One more feature that Knox has, but this is missing is the ability to sleep before retries. A simple flat value like Knox has should be enough.
> 
> https://docs.cloudera.com/runtime/7.2.9/knox-authentication/topics/security-knox-configure-ha.html
Done.

stoty:
Functionally,this looks good.
Some remaining issues:

- As Julien mentioned, URLs including commas are problematic. I think that specifying the encoded comma %2C in the URLs should work. Can you test and conform this ?
- Need to add the new features to the docs
- Need a release note that summarizes the feature (in this JIRA)
- Invite another round of reviews by replying to your original email on the dev list.

vaijosh:
> Functionally,this looks good. Some remaining issues:
> 
> * As Julien mentioned, URLs including commas are problematic. I think that specifying the encoded comma %2C in the URLs should work. Can you test and conform this ?
> * Need to add the new features to the docs
> * Need a release note that summarizes the feature (in this JIRA)
> * Invite another round of reviews by replying to your original email on the dev list.

Thanks @stoty.
I tried %2C but I get URL string like "http://vjoshi21-2.vj%2Coshi2.root.hwx.site:8765". 
Will it be a make sense to add one more property called "LB_URL_SEPARATOR_CHAR" and fix this issue? 
I am little concerned about the number of properties we are exposing with this features. So please advise.

stoty:
That's fine. 
Commas are not valid in host part anyway, only in the path and query
The idea is that the %2c is passed verbatim to the remote server, and then it decodes it.

https://host1:8765/context%2ccontext,https://host2:8765/context%2ccontext
And then Jetty on the server side converts the latter parts back to
"context,context" before processing further

site/_docs/client_reference.md
```diff
@@ -193,3 +193,52 @@ on-hover images for the permalink, but oh well.
...
+: _Default_: `null`.
+
+: _Required_: No.
+
+<strong><a name="lb_strategy" href="#lb_strategy">lb_strategy</a></strong>
+
+: _Description_: The load balancing strategy to be used by the client side load balancer. It must be a fully qualified
+Java class name which implements `org.apache.calcite.avatica.ha.LBStrategy`. Out of the box we have provided two
+implementations `org.apache.calcite.avatica.ha.RandomSelectLBStrategy` and
+`org.apache.calcite.avatica.ha.RoundRobinLBStrategy` which you can use as load balancing strategies.
```
stoty, line 218: nit:
The load balancing strategy to be used by the client side load balancer. It must be a fully qualified
Java class name which implements `org.apache.calcite.avatica.ha.LBStrategy`.
Two implementations are provided: `org.apache.calcite.avatica.ha.RandomSelectLBStrategy` and
`org.apache.calcite.avatica.ha.RoundRobinLBStrategy` .

stoty (changes requested):
Do you have a specific reason for choosing random as the default strategy ?
Knox uses round Robin (only) , and my impression is that round robin tends to be default for most balancers.

site/_docs/client_reference.md
vaijosh: Yes. Here are my thoughts.
Knox is centralized (all client send requests to KNOX) mechanism where as we implemented distributed (all client independently do the LB). For centralized mechanism, RoundRobin makes more sense. But for distributed approach Random select makes more sense. 
To give an example, if we have some clients C1, C2, C3 interacting with DB at the same time using DB URLs DB1,DB2,DB3. So if the clients are doing transactions concurrently it's likely that they can overload Perticular DB server ( e.g. all C1, C2, C3 pickup first DB URL and connect to i). With Random RandomSelect algorithm there is less probability that all clients connect to same DB and overload it.

vaijosh:
> Do you have a specific reason for choosing random as the default strategy ? Knox uses round Robin (only) , and my impression is that round robin tends to be default for most balancers.

Yes this is intentional. Here are my thoughts.
Knox is centralized (all client send requests to KNOX) mechanism where as we implemented distributed (all client independently do the LB). For centralized mechanism, RoundRobin makes more sense. But for distributed approach Random select makes more sense. 
To give an example, if we have some clients C1, C2, C3 interacting with DB at the same time using DB URLs DB1,DB2,DB3. So if the clients are doing transactions concurrently it's likely that they can overload Perticular DB server ( e.g. all C1, C2, C3 pickup first DB URL and connect to i). With Random RandomSelect algorithm there is less probability that all clients connect to same DB and overload it.

site/_docs/client_reference.md
vaijosh: Done. Thanks.

stoty:
What is your opinion on adding a RR variant that starts on a randomly selected host ?

stoty:
Or an RR variant that shuffles the list on initialization ?
I think that that would have the advantage being deterministic, while avoiding the uneven load problem.

vaijosh:
> Or an RR variant that shuffles the list on initialization ? I think that that would have the advantage of being deterministic, while avoiding the uneven load problem.

Interesting proposal. It's better than current RoundRobin.In fact I am tempted to modify RoundRobinLBStrategy and have this. For clients we can expect deterministic behavior ( i.e. which host will come next).  But for servers, it's still an uneven load(Since clients are dependent on Randomiser).
I feel "client side load balancers" can't solve uneven load balancing unless they have integration with some external Discovery service.

stoty:
I suggest keeping RoundRobinLBStrategy as is, and adding something like ShuffledRoundRobinLBStrategy as a subclass.
We could make ShuffledRoundRobinLBStrategy the default, but users can still choose classic RR if they want to.

vaijosh:
> I suggest keeping RoundRobinLBStrategy as is, and adding something like ShuffledRoundRobinLBStrategy as a subclass. We could make ShuffledRoundRobinLBStrategy the default, but users can still choose classic RR if they want to.

Done

stoty (changes requested):

core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+  Map<String, String[]> configToUrlListMap = new HashMap<>();
+
+  @Override
+  public String getLbURL(ConnectionConfig config) {
+    String key = getKey(config);
+    int urlIndex;
+    String lbURLs = config.getLbURLs();
+    if (!configToIndexServedMap.containsKey(key)) {
+      String[] urls = lbURLs.split(URL_SEPERATOR_CHAR);
+      configToIndexServedMap.put(key, new Random().nextInt(urls.length));
```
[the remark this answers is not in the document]
vaijosh: @stoty 
If we shuffle the URLs it will degenerate to RandomSelectLBStrategy.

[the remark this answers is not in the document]
stoty: I disagree.
If we shuffle the list before storing it into the hash, then each VM locally will still behave deterministically, but each one will still behave differently.

for the URLs ABCD

JVM1:
RR on CBDA

JVM2:
RR on DCBA

JVM3:
RR on ADCB

and so on.

With true random strategy you will get something like ACCDBADD.... where you can re-try a failed host immediately.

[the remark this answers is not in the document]
vaijosh: Thanks @stoty 
Yes. I got your point. So shuffle will solve repetition problem within JVM, so its better than RandomSelect in this sense. I will make these changes tomorrow.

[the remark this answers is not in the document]
vaijosh: Done

core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
```diff
@@ -0,0 +1,74 @@
...
+    String url = urls[urlIndex];
+    LOG.info("Selected URL:{}", url);
+    return url;
+  }
+
+  private void initialiseCaches(String key, String lbURLs) {
+    String[] urls = lbURLs.split(URL_SEPERATOR_CHAR);
+    List<String> list = Arrays.asList(urls);
+    Collections.shuffle(list);
+    list.toArray(urls);
```
stoty, line 66: this doesn't do anything.

core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
```diff
@@ -0,0 +1,74 @@
...
+  @Override
+  public String getLbURL(ConnectionConfig config) {
+    String key = getKey(config);
+    String lbURLs = config.getLbURLs();
+    if (!configToIndexServedMap.containsKey(key)) {
+      initialiseCaches(key, lbURLs);
+    }
+    String[] urls = configToUrlListMap.get(key);
+    int urlIndex = configToIndexServedMap.get(key);
+    int nextURlIdx = urlIndex < urls.length - 1 ? (urlIndex + 1) : 0;
```
stoty, line 55: nit:
Use inc + modulo, that's easier to read.

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  public String getLbURL(ConnectionConfig config) {
+    String key = getKey(config);
+    if (!configToIndexServedMap.containsKey(key)) {
+      configToIndexServedMap.put(key, 0);
+      configToUrlListMap.put(key, config.getLbURLs().split(URL_SEPERATOR_CHAR));
+    }
+    String[] urls = configToUrlListMap.get(key);
+    int urlIndex = configToIndexServedMap.get(key);
+    int nextURlIdx = urlIndex < urls.length - 1 ? (urlIndex + 1) : 0;
```
stoty, line 51: nit:
Use inc + modulo, that's easier to read.

core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
```diff
@@ -0,0 +1,74 @@
...
+    configToIndexServedMap.put(key, nextURlIdx);
+    String url = urls[urlIndex];
+    LOG.info("Selected URL:{}", url);
+    return url;
+  }
+
+  private void initialiseCaches(String key, String lbURLs) {
```
stoty, line 62: nit:
This isn't really a cache. Better call this something else, like initialiseUrlList

stoty (changes requested):

core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
vaijosh: @stoty 
It works as follows
1. Convert from array to list using Arrays.asList()
2. Collection.shuffle shuffles the list passed as an argument ( void return)
3. List.toArray( array ) - Put the shuffled elements in original array. [Check this](https://devdocs.io/openjdk~8/java/util/list). We  can ignored the return because we know that passed array will have shuffled elements.
Please let me know if you see any issue in above behavior or other suggestion.

stoty: You are correct.
However, relying on that side effect hurts readability.
`list = Collections.shuffle(list);`
is obviously correct, and does the same thing.

core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
vaijosh: Done. Made it more readably.

core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
vaijosh: Done.

core/src/main/java/org/apache/calcite/avatica/ha/ShuffledRoundRobinLBStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+      configToIndexServedMap.put(key, 0);
+      initialiseUrlList(key, lbURLs);
+    }
+    String[] urls = configToUrlListMap.get(key);
+    int urlIndex = configToIndexServedMap.get(key);
+
+    String url = urls[urlIndex];
+    LOG.info("Selected URL:{}", url);
+
+    urlIndex++;
```
stoty, line 60: I had 

urlIndex = (urlIndex + 1) % urls.length

or

urlIndex = ++urlIndex % urls.length

in mind

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
vaijosh: Done

stoty (approved):
+1 LGTM

stoty:
Looks good to me, but it would be nice to get another review before committing.
@joshelser Do you have the time take a look ?

joshelser:
Sorry, I was out last week. Looking :)

site/_docs/client_reference.md
```diff
@@ -193,3 +193,52 @@ on-hover images for the permalink, but oh well.
...
+load balancer. Depending on the load balancing strategy, load balancer selects one of the URLs from the list.
+
+: _Default_: `null`.
+
+: _Required_: No.
+
+<strong><a name="lb_strategy" href="#lb_strategy">lb_strategy</a></strong>
+
+: _Description_: The load balancing strategy to be used by the client side load balancer. It must be a fully qualified
+Java class name which implements `org.apache.calcite.avatica.ha.LBStrategy`. Two implementations are provided
```
joshelser, line 216: "Three" implementations appear to be provided.

core/src/main/java/org/apache/calcite/avatica/ha/RandomSelectLBStrategy.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Random;
+
+/**
+ * Random Select strategy for client side load balancing.
+ */
+public class RandomSelectLBStrategy implements LBStrategy {
```
joshelser, line 29: If you choose a random Avatica server every time, the client will (effectively) be stuck in a retry loop to re-create new Connection and Statements on that Avatica server, to then be bounced to a different Avatica server.

I think this implementation would be overtly harmful should a user ever try to configure it.

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Round Robin strategy for client side load balancing.
+ * Its implemented it as a singleton so that we can maintain state
+ * i.e. which URL was last used from the list of URLs specified.
+ */
+public class RoundRobinLBStrategy implements LBStrategy {
```
joshelser, line 32: Same with this implementation, if the caller is invoking a different Avatica server every call, the client will be "stuck" constantly trying to create and use the Statement/Connection which were on the n-1th server.

joshelser (changes requested):
Unless I've missed something quite fundamental, none of these client-side load balancers will actually be effective to run queries. The Avatica server is not completely stateless because JDBC itself is not stateless. We have to maintain Connections and Statements in memory, proxying the calls on those JDBC objects from Avatica client to Avatica server.

The Avatica client will try to do some work to re-create a Connection or Statement which is not present on the Avatica server, but this is meant to be an infrequent task (e.g. if the Avatica server restarted). If paging through a ResultSet is interrupted, the Avatica server has no choice but to fail hard because it cannot pick up where the previous reader left off.

A long time ago, I wrote this article https://community.cloudera.com/t5/Community-Articles/Deploying-the-Phoenix-Query-Server-in-production/ta-p/245104 about how the only reasonable way to implement load balancing safely is to use sticky sessions.

I'm of the mindset that it's better to leave the Avatica client as "dumb", and implement HTTP load balancing with stickiness using projects whose purpose is _just_ that, e.g. nginx, httpd, haproxy, etc. I'm not -1 on making the Avatica client do some of the load balancing itself, but I don't think this PR does what you expect it to do (hopefully I'm mistaken).

I'd suggest you write some unit tests that exercise one of your load balancers and actually run some queries against hsqldb.

On a good note, thank you for including documentation updates with your change. That's a great sight to see.

core/src/main/java/org/apache/calcite/avatica/ha/RandomSelectLBStrategy.java
vaijosh: Thanks @joshelser  for the inputs.

For internal failover retries ( done by driver) we have "lb_connection_failover_retries" to s address this issue. By default we will do only 3 failover retries after connection will be marked as failed. 

Not sure if you are pointing about some scenarios related to external retries( implemented by application).

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
vaijosh: As I mentioned above "lb_connection_failover_retries" should address this concern.

stoty:
The automatic reconnect feature has been disabled by default in CALCITE-5009, @joshelser .
Adding a check so that the LB feature and the automatic reconnect cannot be enabled together is a good idea, though, and we should do that, @vaijosh.

This is how it's supposed to work:

With CALCITE-5009, there is a strict one-to-one mapping between JDBC connection and HTTP client connection objects, and there is no way that a new HTTP Client connection object pointing to different URL can get associated with a JDBC connection during its lifetime.

The URL in the HTTP client connection object for a JDBC connection doesn't change after the JDBC Connection has been initialized. A new JDBC Connection is only created ( in Driver#connect() ) when the user explicitly requests a new Connection object from JDBC. The LB host selection / retery  code is guaranteed to only ever run before the user receives a new Connection, and never on an active JDBC Connection.
This ensures stickyness, if the server goes away, then the JDBC Connection simply fails normally, and the application has to create a new one.

Of course, we might have missed some code path where that happens, but this is the general idea.

Having a test specifically for this behaviour would be a good thing. I'm thinking of creating a single connection, and issuing BEGIN, upsert, COMMIT sequence from three different statements created from the same connection would be a good way to demonstrate that they all connect to the same Avatica server.
We could also add a test that checks wich Avatica server an active connection is using, kills that server, and the checks that the connection goes into a fail state without trying to transparently re-connect.

As for using an external LB solution, the main motivation for this patch was Kerberos/SPNEGO, which precludes the use of standard external LBs.
Apart from removing the need for Knox (the only LB that works with Kerberos that I know), adding the LB code into the client also removes the LB as a single point of failure. (Which otherwise requires additional DNS or network level complexity)
We also haven't tested yet whether the new Jetty kerberos code that you've enabled in Avatica fixes all the known perf problems with Knox. I suspect that it fixes well-behaved applications that reuse the JDBC connection object, and hence the HTTP session, but spamming new Connections for each statement would stil overload the identity infra.

stoty:
Oh and thanks for taking the time to look at this, @joshelser .

stoty:
Thinking some more about this, LB doesn't make the transaprent reconnect sitation worse than it is without LB, it doesn't really matter if we re-create an unrelated new connection on the same server, or on a different server, the state is lost just the same.

site/_docs/client_reference.md
vaijosh: Thanks @joshelser.
I have changed it.

vaijosh:
> 

I have added a test for a following scenario.

1. Start few server.
2. Create connection
3. Specify the client side LB properties with RoundRobinLBStrategy.
4. Stop the server which appears first in the list if lbURLs
5. Try to create statement. Here we make sure that connection is failed ( and not attempted with different URL).

server/src/test/java/org/apache/calcite/avatica/ha/ConnectionPropertiesHATest.java
```diff
@@ -0,0 +1,311 @@
...
+    sb.append("http://localhost:").append(test_server_port).append(",");
+    for (int i = 0; i < 1000; i++) {
+      sb.append("http://localhost:").append(START_PORT + i).append(",");
+    }
+    properties.put(BuiltInConnectionProperty.LB_URLS.name(), sb.substring(0, sb.length() - 1));
+
+    //Create a connection
+    String url = SERVERS.getJdbcUrl(test_server_port, Driver.Serialization.PROTOBUF);
+    AvaticaConnection conn = (AvaticaConnection) DriverManager.getConnection(url, properties);
+
```
stoty, line 278: Better have a test here first here, to make sure that the connection works before we shut the server down.
(Just create a connection, and select anything)

stoty: Actually, to exercise the most code paths, and make sure we do not transparently reconnect:
Create connection1
Create statement1
Execute select
Kill server
Try to execute a query on statement1 (expect to fail)
Try to create statement2 on connection1(expect to fail)

vaijosh: Done

stoty (approved):
Thanks, the new test illustrates the sticky behaviour well.

joshelser:
> With [CALCITE-5009](https://issues.apache.org/jira/browse/CALCITE-5009), there is a strict one-to-one mapping between JDBC connection and HTTP client connection objects, and there is no way that a new HTTP Client connection object pointing to different URL can get associated with a JDBC connection during its lifetime.

Ah, I'm too far removed! I'll try to re-read today and think through this some more. Thanks for the great explanation. I was quite confused because it seemed like there was something fundamentally missing -- turns out what was missing was my understanding :)

vaijosh:
> 
Hi @joshelser 
Did you get a chance to review this?

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
```diff
@@ -40,18 +40,15 @@ private RoundRobinLBStrategy() { }
   Map<String, String[]> configToUrlListMap = new HashMap<>();
 
   @Override
-  public String getLbURL(ConnectionConfig config) {
+  public synchronized String  getLbURL(ConnectionConfig config) {
```
stoty, line 43: Nit double space

core/src/main/java/org/apache/calcite/avatica/remote/Driver.java
```diff
@@ -194,10 +196,20 @@ AvaticaHttpClient getHttpClient(AvaticaConnection connection, ConnectionConfig c
       } catch (Exception e) {
         long endTime = System.currentTimeMillis();
         long elapsedTime = endTime - startTime;
+        LOG.warn("Connection Failed: {}", e.getMessage());
         LOG.debug("Failure detected in: {} milliseconds", elapsedTime);
-        if (currentRetry < retries) {
+        if (currentRetry < retries - 1) {
```
stoty, line 201: This means that if you set retries to 1, then you won't retry at all.
Technically, retry = 1 mens and initial attempt, and the one retry.
IMO the previous version was correct,

stoty: The docs you wrote also says that this the number of REtries.

stoty (changes requested):

core/src/main/java/org/apache/calcite/avatica/remote/Driver.java
vaijosh: Done.

core/src/main/java/org/apache/calcite/avatica/ha/RoundRobinLBStrategy.java
vaijosh: Done

stoty (approved):
+1 LGTM

vaijosh:
> One more thing, @vaijosh . Please change the commit message to match the Avatica convention. i.e. use square brackets instead of colon.
Done. Also squashed 2 commits into single.

stoty:
The commit message doesn't match the JIRA description
please use the JIRA description verbatim.

vaijosh:
> 

Done

stoty (changes requested):
The docs formatting for the new options is off. (defaults are not formatted as for the rest of the options)
You can preview it by following the instructions in site/README.md
(I had to replace "docker compose" with "docker-compose")

stoty (changes requested):
The docs formatting for the new options is off. (defaults are not formatted as for the rest of the options)
You can preview it by following the instructions in site/README.md
(I had to replace "docker compose" with "docker-compose")

vaijosh:
> The docs formatting for the new options is off. (defaults are not formatted as for the rest of the options) You can preview it by following the instructions in site/README.md (I had to replace "docker compose" with "docker-compose")
Thank @stoty 
I have fixed the formatting.

stoty (approved):
+1 LGTM

stoty:
The commit message doesn't include the JIRA number, but I will fix that on commit.

stoty:
merged manually

PR #220, opened 2023-06-03
[CALCITE-5748] Guava 32.1.1-jre

snuyanzin:


richardantal:
I wonder if there is a reason why we don't go to the latest/greatest 32.1.1?

snuyanzin:
the PR was created when only 32.0.0 was present, that's the reason
i will update it

richardantal:
Looks good to me, Thank you Sergey for the change.
I don't have permissions to review, maybe @stoty does.

stoty (approved):
+1 LGTM

snuyanzin:
Thanks for having a look and sorry for the delay

snuyanzin:
rebased and squashed commits

PR #224, opened 2023-08-03
[CALCITE-5890] Handle non-JKS truststores in Avatica client

stoty:


stoty:
Successfully tested on a FIPS compliant JVM with the BouncyCastle provider.

stoty:
Note to self:
The local branch neme is wrong. 
(This does not affect the contents of the PR.)

stoty:
Added docs.

snuyanzin:
Thanks for the PR
it looks ok from my side
The only thing I wonder is whether there is an easy way to cover it with tests?

stoty:
There is no easy way to do that.
Without a lot of effort, I could add a test where I specify "JKS" (the default), and another one where I specify a non-existent one, and check that that one fails.
This would at least test that setting does apply.

stoty:
I have added a test @snuyanzin.

It's not a full end to end crypto test, but it does cover the few new lines where we configure the HTTP client object.

It turns out that this test required a change in the SSL context equivalency check code.
This wouldn't come up in the real world, but won't hurt either.

stoty:
Is this test sufficient, @snuyanzin  ?

snuyanzin (approved):
>It's not a full end to end crypto test, but it does cover the few new lines where we configure the HTTP client object.

Thanks for adding tests, I think it is good that it covers configuration of HTTP client

PR #225, opened 2023-08-29
[CALCITE-5957] Date representation string without precede 0 is acceptable

zstan:


core/src/test/java/org/apache/calcite/avatica/util/TimestampFromNumberAccessorTest.java
```diff
@@ -101,6 +101,9 @@ public class TimestampFromNumberAccessorTest {
     value = DateTimeUtils.timestampStringToUnixDate("2014-09-30 15:28:27.356");
     assertThat(instance.getString(), is("2014-09-30 15:28:27"));
 
+    value = DateTimeUtils.timestampStringToUnixDate("2014-2-2 15:28:27.356");
```
herunkang2018, line 104: Could you add test case to cover `'1945-2-2 12:2:5'` and `'12:2:5' to test the TIME type without leading zero?

zstan: @herunkang2018 done.

core/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java
```diff
@@ -61,11 +61,11 @@ private DateTimeUtils() {}
 
   /** Regex for date, YYYY-MM-DD. */
   private static final Pattern ISO_DATE_PATTERN =
-      Pattern.compile("^(\\d{4})-([0]\\d|1[0-2])-([0-2]\\d|3[01])$");
+      Pattern.compile("^(\\d{4})-([0]?\\d|1[0-2])-([0-2]?\\d|3[01])$");
 
   /** Regex for time, HH:MM:SS. */
   private static final Pattern ISO_TIME_PATTERN =
-      Pattern.compile("^([0-2]\\d):[0-5]\\d:[0-5]\\d(\\.\\d*)*$");
+      Pattern.compile("^([0-2]\\d):[0-5]?\\d:[0-5]?\\d(\\.\\d*)*$");
```
herunkang2018, line 68: Is `'1:2:5'` also a valid time, with the hour without leading zero? IMO yes, what do you think?

zstan: @herunkang2018 ok, done.

herunkang2018:
@zstan LGTM

PR #226, opened 2023-09-07
[CALCITE-5981] TIMESTAMPDIFF function returns incorrect result

mihaibudiu:
I have cross-checked the implementation with the MySQL implementation:
https://github.com/mysql/mysql-server/blob/ea1efa9822d81044b726aab20c857d5e1b7e046a/sql/item_timefunc.cc
The tests are all adapted from MySQL, where they look like: "timestampdiff(month, ...)"
https://github.com/mysql/mysql-server/blob/ea1efa9822d81044b726aab20c857d5e1b7e046a/mysql-test/r/func_time.result#L1130


asolimando (approved):
LGTM, will merge in a couple of days unless there are further comments.

PR #227, opened 2023-09-11
[CALCITE-5964] Add test case checking that connection.getMetaData().getTables() handles responses with different signature shapes.

olivrlee:
Also adding in Type.CLASS to ConnectionProperty.java and the associated getters.


core/src/main/java/org/apache/calcite/avatica/MetaImpl.java
```diff
@@ -388,6 +388,32 @@ public MetaColumn(
     public String getName() {
       return columnName;
     }
+    public static String[] getColumnNames() {
```
wnob, line 391: Might as well link to some documentation: https://docs.oracle.com/en/java/javase/20/docs/api/java.sql/java/sql/DatabaseMetaData.html#getColumns(java.lang.String,java.lang.String,java.lang.String,java.lang.String)

Also add a comment explaining why anybody might want to override this.

core/src/main/java/org/apache/calcite/avatica/remote/MockJsonService.java
```diff
@@ -88,6 +88,82 @@ public Service create(AvaticaConnection connection) {
...
+                + "\"definitelyWritable\": false,\"columnClassName\": \"java.lang.String\"}],\"sql\":null,"
+            + "\"parameters\":[],"
+            + "\"cursorFactory\":{\"style\":\"LIST\",\"clazz\":null,\"fieldNames\":null},\"statementType\":null},"
+            + "\"firstFrame\":{\"offset\":0,\"done\":true,"
+            + "\"rows\":["
+              + "[\"PUBLIC\",\"SCOTT\",\"DEPT\",\"TABLE\",null,null,null,null,null,null,\"MEMORY\",false,null,\"EXTRA_LABEL1\"],"
+              + "[\"PUBLIC\",\"SCOTT\",\"EMP\",\"TABLE\",null,null,null,null,null,null,\"MEMORY\",false,null,\"EXTRA_LABEL2\"],"
+              + "[\"PUBLIC\",\"SCOTT\",\"BONUS\",\"TABLE\",null,null,null,null,null,null,\"MEMORY\",false,null,\"EXTRA_LABEL3\"],"
+              + "[\"PUBLIC\",\"SCOTT\",\"SALGRADE\",\"TABLE\",null,null,null,null,null,null,\"MEMORY\",false,null,null]]},"
+            + "\"updateCount\":-1,\"rpcMetadata\":null}"
```
wnob, line 161: This JSON blob is impenetrable. How did you produce this?

server/src/test/java/org/apache/calcite/avatica/RemoteDriverMockTest.java
```diff
@@ -127,6 +127,21 @@ private Connection getMockConnection() {
     connection.close();
   }
 
+  @Test public void testTablesAdditionalColumns() throws Exception {
```
wnob, line 130: I'm not convinced of the usefulness of this test. Which part of the production code is actually being exercised? Kinda looks like we're just stubbing a mock, then testing that we stubbed it.

core/src/main/java/org/apache/calcite/avatica/remote/MockJsonService.java
olivrlee: I ran another test within Avatica and took that json blob and modified it 
What does 'impenetrable' mean here?

server/src/test/java/org/apache/calcite/avatica/RemoteDriverMockTest.java
olivrlee: It mocks getting the response, but it actually tests the client code in RemoteMeta that handles the response and parses the signature into the ResultSet.

core/src/main/java/org/apache/calcite/avatica/MetaImpl.java
barrkel: It's static so it can't be overridden, the reader thinks. The fact that the caller is using reflection to find and invoke it could definitely do with some commentary, if that is indeed the best way for the caller to get this information.

barrkel, line 391: Why `String[]` instead of `List<String>`? Why not return a `Collections.unmodifiableList` stored in a static readonly field? This news up a list and three arrays (including one for the varargs) and does two copies every time it's called, absent JVM optimizations. Not the end of the world, but it's easy to reduce the overhead if you can use Lists instead of arrays, as Effective Java and almost everyone else advises.

core/src/main/java/org/apache/calcite/avatica/remote/MockJsonService.java
barrkel: It's hard to read this data fixture and understand its direct relationship with assertions in tests.

There's also the usual problem with mocking - how do you keep your mocks in sync with the real implementation. When Avatica changes its response, tests depending on this fixture will go stale without further notice and there is no forcing function to cause them to be updated until or unless a regression occurs. But this problem is hard to solve without designing for it.

core/src/main/java/org/apache/calcite/avatica/MetaImpl.java
olivrlee: The reason why I kept that was because of how `createResultSet` is set up to take in the varags.
I could restructure `createResultSet` to work with `List<String>`

What would be the best way to instantiate this when considering the number of copies?

How is: 
```
public static final List<String> COLUMN_NAMES = Collections.unmodifiableList(
      Arrays.asList("TABLE_CAT",
          "TABLE_SCHEM",
          "TABLE_NAME",
          "TABLE_TYPE",
          "REMARKS",
          "TYPE_CAT",
          "TYPE_SCHEM",
          "TYPE_NAME",
          "SELF_REFERENCING_COL_NAME",
          "REF_GENERATION"));

Comment (barrkel) [reply core/src/main/java/org/apache/calcite/avatica/MetaImpl.java:391]: So if it's most ergonomic as a string array, then just new up a string array directly rather than going through asList -> toArray. As in `return new String[] { "TABLE_CAT", "TABLE_SCHEM", ... };`

File: core/src/main/java/org/apache/calcite/avatica/remote/MockJsonService.java (hunk 2 above)
Comment (olivrlee) [reply core/src/main/java/org/apache/calcite/avatica/remote/MockJsonService.java:161]: So one thing to note is that this test really just shows that whatever Signature comes in (different columns, different orders), Avatica is able to handle the different shapes and read them in the supplied order.
It doesn't test whether the real implementation is providing a correct response in the right order. That part should be up to the implementation. 

Maybe that point wasn't clear by this additional test, perhaps I should shorten the JSON to contain fewer columns and add a comment stating that the point is that the signature column order matches up with the resultset columns?

File: core/src/main/java/org/apache/calcite/avatica/ConnectionConfigImpl.java [hunk 4]
```diff
@@ -337,6 +337,19 @@ public <T> T getPlugin(Class<T> pluginClass, String defaultClassName,
       return get_(pluginConverter(pluginClass, defaultInstance),
           defaultClassName);
     }
+
+    /** Returns the class value of this property. Throws if not set and no
+     * default. */
+    public Class<?> getClazz(Class<?> clazz, Class<?> defaultValue) {
```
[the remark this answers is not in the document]
julianhyde: There's a small chance for confusion with `Object.getClass()`. So, I would recommend calling it `getType`.

PR #228, opened 2023-10-02
[CALCITE-5957] Date and time representation string without precede 0 is acceptable

herunkang2018:


core/src/test/java/org/apache/calcite/avatica/util/TimestampFromNumberAccessorTest.java
```diff
@@ -101,6 +101,24 @@ public class TimestampFromNumberAccessorTest {
...
+    value = DateTimeUtils.timestampStringToUnixDate("1945-2-2 2:2:5");
+    assertThat(instance.getString(), is("1945-02-02 02:02:05"));
+
+    value = DateTimeUtils.timestampStringToUnixDate("900-2-2 2:2:5");
+    assertThat(instance.getString(), is("0900-02-02 02:02:05"));
+
+    value = DateTimeUtils.timestampStringToUnixDate("9-2-2 2:2:5");
+    assertThat(instance.getString(), is("0009-02-02 02:02:05"));
+
+    value = DateTimeUtils.timestampStringToUnixDate("2023-01-01 22:00:00.123.123");
```
herunkang2018, line 119: This test case is not a regression, it is passed before and after CALCITE-5678. We add it here for better code coverage.

stoty:
Copíing the last comment on the ticket here:

> [Stamatis Zampetakis](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=zabetak) added a comment - [03/Oct/23 11:39](https://issues.apache.org/jira/browse/CALCITE-5957?focusedCommentId=17771389&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17771389)
Based on the current discussion under https://lists.apache.org/thread/fvh4nfbm9mvt9dfpo6o8nq37q6mmgqh0 there is no strong desire to restore the old behavior so let's leave this ticket open for now and consider that the expected and correct behavior from now on is the one introduced by [CALCITE-5678](https://issues.apache.org/jira/browse/CALCITE-5678).

PR #232, opened 2023-11-27
[CALCITE-6135] BEARER authentication support

Aarchy:


core/src/main/java/org/apache/calcite/avatica/remote/SingleBearerTokenProvider.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package org.apache.calcite.avatica.remote;
+
+import org.apache.calcite.avatica.BuiltInConnectionProperty;
+import org.apache.calcite.avatica.ConnectionConfig;
+
+import java.io.IOException;
+
+public class SingleBearerTokenProvider implements BearerTokenProvider {
```
stoty, line 24: That name sounds off. Every provider provides a single token (at a time).
Wouldn't SimpleBearerTokenProvider be more descriptive ?

Aarchy: This object can provide a single token for its entire lifetime. Maybe Constant?
I'd avoid "Simple" as it may sound that the token is simple in some way.

stoty: Constant is better.

stoty:
Can you open a [DISCUSS] thread on the calcite list about this change ?
Explain its background in some detail ?

Aarchy:
@stoty Based on your suggestion on the devlist, token providers now can be plugged in via config properties. Users can define their custom properties and use it in token provider's init function. There's a very basic example in the tests.
On the other hand, the current way of handling properties is a bit confusing in my opinion. For now, I tried to just modify the necessary parts.

stoty:
For the record:
This PR copies some of the Bear auth code from httpclient.
What we should do instead do is upgrading to a httpclient with native bearer auth support.

Either to 5.5, which is hopefully going to a have a new SPNEGO implementation, or to 5.4, and re-enable the deprecated SPENGO authentication in it..

stoty:
This needs heavy refactoring, now that we have updated httpclient to one which has native BEARER support.
(Copied code needs to be removed)

PR #233, opened 2023-11-27
[CALCITE-6141] Add jdk8.checkstyle property, use jdk8.checkstyle in case of java 8

snuyanzin:
The idea is to use `jdk8.checkstyle.version` gradle property which will be used in case of java 8
for other java version `checstyle.version` will be used
This allows to compile Avatica with java 8 without explicitly set checkstyle version

also mentioned at https://lists.apache.org/thread/r27kdvvhr0222nqlog7xzn5y3yjvz6jp

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,7 +58,7 @@ jobs:
           fetch-depth: 50
       - name: 'Test'
         run: |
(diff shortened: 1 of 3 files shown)


F21:
@snuyanzin Can you open a jira issue for this and update the commit message?

snuyanzin:
@F21 yes, sure, jira issue is created CALCITE-6141, also commit message changed

F21 (approved):

PR #238, opened 2024-02-09
[CALCITE-6248] Illegal dates are accepted by casts

mihaibudiu:


mihaibudiu:
@zstan this builds on top of your previous fix for a similar issue with timestamps.

core/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java
```diff
@@ -739,6 +744,25 @@ private static int parseFraction(String v, int multiplier) {
     return r;
   }
 
+  /** Lenient date validation.  This accepts more date strings
+   * than validateDate: it does not insist on having two-digit
+   * values for days and months, and accepts spaces around the value.
+   * @param s     A string representing a date.
+   */
+  private static void validateLenientDate(String s) {
```
YiwenWu, line 752: There is no problem using this method to verify, but this implementation only verifies `DAY`. Is it possible to use a more general verification method, such as：

```
DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
try {
    LocalDate date = LocalDate.parse(dateString, formatter);
    return true;
} catch (DateTimeParseException e) {
    return false;
}
```

mihaibudiu: This does the exact same validation as the existing function validateDate.
So maybe we should improve that function too.

mihaibudiu: I have added a commit to validate month and year values.
Turns out year 0 is not legal in the Gregorian calendar, so I had to delete some tests that were using it.

YiwenWu (approved):

mihaibudiu:
If CI passes I will merge this PR.

PR #241, opened 2024-02-27
[CALCITE-6282] Avatica ignores time precision when returning TIME results

mihaibudiu:
…

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
@@ -154,7 +154,7 @@ public abstract class AbstractCursor implements Cursor {
       case PRIMITIVE_INT:
       case INTEGER:
       case NUMBER:
-        return new TimeFromNumberAccessor(getter, localCalendar);
+        return new TimeFromNumberAccessor(getter, localCalendar, columnMetaData.precision);
       case JAVA_SQL_TIME:
         return new TimeAccessor(getter, localCalendar);
       default:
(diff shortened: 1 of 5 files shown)


jduo (approved):

core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -955,10 +955,12 @@ protected Number getNumber() throws SQLException {
    */
   static class TimeFromNumberAccessor extends NumberAccessor {
     private final Calendar localCalendar;
+    private final int precision;
 
-    TimeFromNumberAccessor(Getter getter, Calendar localCalendar) {
+    TimeFromNumberAccessor(Getter getter, Calendar localCalendar, int precision) {
       super(getter, 0);
```
YiwenWu, line 961: I have a question, should the 2 in `Time(2)` be `scale` or `precision`?

core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -1220,7 +1227,7 @@ static class TimestampAccessor extends ObjectAccessor {
       }
       final long unix =
           DateTimeUtils.sqlTimestampToUnixTimestamp(timestamp, localCalendar);
-      return timestampAsString(unix, null);
+      return timestampAsString(unix, null, this.precision);
```
YiwenWu, line 1230: For `cast(TIME '12:42:25.34' as TIME(2))` case,  whether to consider 
 `DateTimeUtils.unixTimeToSqlTime`  corresponding changes

core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
mihaibudiu: I will rename it to "scale"

mihaibudiu: I guess I have to leave it as `precision`, since both the columnMetadataFields which originate this value call it `precision`, and the helper functions that manipulate this field also call it `precision` (as you point out in your other comment).  It would be too confusing to use different names for this value in different parts of the codebase. 

I propose we file a new issue if we want to modify the field name to `scale`, to make sure that we cover all the software layers involved. In the meantime it would be great to merge this PR because it fixes a genuine bug.

YiwenWu (approved):

mihaibudiu:
So what's the plan about merging this PR? Do I have to make somehow the CI work?

F21:
I would suggest we temporary comment out the tests in Calcite and renable them when it's upgrade to Avatica 1.25.0. See Julian's suggestion: https://lists.apache.org/thread/b7kb393yvj10rbmsxv4bh076bk2ncx58

PR #246, opened 2024-08-14
[CALCITE-6530] HTTP Sessions are never expired in Avatica server

stoty:


joshelser (approved):
Was reading a stackoverflow post and it looks like Jetty defaults this to 30mins when using their web descriptor (which Avatica isn't): https://stackoverflow.com/questions/49300484/how-to-set-the-http-session-timeout-in-jetty-9-embedded

Looks good to me (but I am distant from the codebase). I can't imagine that this needs to be user-configurable, either.

stoty:
I have read the same article :)
Yes, I came to the same conclusion regarding the web descriptor and the default idle timeout.

stoty:
Thanks for the quick reply Josh.

joshelser:
> I have read the same article :) Yes, I came to the same conclusion regarding the web descriptor and the default idle timeout.

Excellent, I figured as much!

> Thanks for the quick reply Josh.

You bet, caught me at a good time and this change looked straightforward 👍🏼

PR #249, opened 2024-09-13
CALCITE-6529: Use persistent sessionContext in AvaticaCommonsHttpClie…

szucsvillo:
…ntImpl


mihaibudiu (approved):
I am not familiar with this part of the codebase, but these changes look reasonable.

mihaibudiu:
The CI failed for macos

caicancai (approved):
Please modify the commit message and PR title to: [CALCITE-6529] xxxxx

server/src/test/java/org/apache/calcite/avatica/server/HttpServerSpnegoWithoutJaasTest.java
```diff
@@ -233,8 +233,8 @@ private static void setupUsers(File keytabDir) throws KrbException {
         // Passes the GSSCredential into the HTTP client implementation
         final AvaticaCommonsHttpClientImpl httpClient =
             new AvaticaCommonsHttpClientImpl(httpServerUrl);
-        httpClient.setGSSCredential(credential);
         httpClient.setHttpClientPool(pool, config);
+        httpClient.setGSSCredential(credential);
```
stoty, line 237: Why swap these ?

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
```diff
@@ -109,34 +110,33 @@ protected void initializeClient(PoolingHttpClientConnectionManager pool,
     HttpClientBuilder httpClientBuilder = HttpClients.custom().setConnectionManager(pool)
         .setDefaultRequestConfig(requestConfig);
     this.client = httpClientBuilder.build();
+
```
stoty, line 113: have you tried setting these directly on the client object ?

szucsvillo: Not yet but i will try.

server/src/test/java/org/apache/calcite/avatica/server/HttpServerSpnegoWithoutJaasTest.java
szucsvillo: Inside setHttpClientPool, the initializeClient method is called, which initializes the persistentContext object. After that setGSSCredential method sets the credentials within persistentContext.
If we try to set the credentials before initializing persistentContext, it would result in a NullPointerException.

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
szucsvillo: I have tried setting these directly on the client object, but I think it would require quite a lot of refactoring. For example, practically the whole AvaticaHttpClientFactoryImpl class should be rewritten.

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
```diff
@@ -91,6 +91,7 @@ public class AvaticaCommonsHttpClientImpl implements AvaticaHttpClient, HttpClie
   protected CredentialsProvider credentialsProvider = null;
   protected Lookup<AuthSchemeFactory> authRegistry = null;
   protected Object userToken;
+  protected HttpClientContext persistentContext;
```
stoty, line 94: I think we can call this simply context or HttpContext.
Persistence should be implied.

core/src/main/java/org/apache/calcite/avatica/remote/AvaticaCommonsHttpClientImpl.java
stoty: Good point.
Maybe in a followup ticket.

stoty:
Please fix the commit message to use square brackets, @szucsvillo .

stoty (approved):
+1 LGTM

PR #250, opened 2024-09-24
CALCITE-6590 Remove use of Java SecurityManager in Avatica

stoty:


stoty:
This lets all tests pass.

Questions:

- Should I split the ByteBuddy upgrade to a separate ticket + patch ?
- I guess we also want to rename the ticket

PR #251, opened 2024-09-26
[CALCITE-6590] Use reflection to handle Java SecurityManager deprecation in Avatica

stoty:
Also bump ByteBuddy version to 1.15.1

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
@@ -156,7 +156,7 @@ public class SecurityUtils {
     try {
       return (T) doPrivileged.invoke(action);
     } catch (Throwable t) {
-      throw inferAndCast(t);
+      throw sneakyThrow(t);
     }
   }
 
@@ -173,7 +173,7 @@ public class SecurityUtils {
     } catch (PrivilegedActionException e) {
       throw new CompletionException(e.getCause());
     } catch (Throwable t) {
-      throw inferAndCast(t);
+      throw sneakyThrow(t);
     }
   }
 
@@ -185,7 +185,7 @@ public class SecurityUtils {
     try {
       return (Subject) CURRENT.invoke();
     } catch (Throwable t) {
-      throw inferAndCast(t);
+      throw sneakyThrow(t);
     }
   }
 
@@ -196,7 +196,7 @@ public class SecurityUtils {
   }
 
   @SuppressWarnings("unchecked")
-  private static <E extends Throwable> RuntimeException inferAndCast(Throwable e) throws E {
+  private static <E extends Throwable> RuntimeException sneakyThrow(Throwable e) throws E {
     throw (E) e;
   }
 }
--- a/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
@@ -32,13 +32,13 @@ import javax.security.auth.Subject;
  * of the security classes defined by <a href="https://openjdk.org/jeps/411">JEP 411</a>.</p>
  */
 public class SecurityUtils {
-  private static final MethodHandle DO_AS = lookupDoAs();
-  private static final MethodHandle DO_PRIVILEGED = lookupDoPrivileged();
-  private static final MethodHandle GET_CONTEXT = lookupGetContext();
-  private static final MethodHandle GET_SUBJECT = lookupGetSubject();
+  private static final MethodHandle CALL_AS = lookupCallAs();
   private static final MethodHandle CURRENT = lookupCurrent();
+  private static final MethodHandle DO_PRIVILEGED = lookupDoPrivileged();
+  private static final MethodHandle GET_SUBJECT = lookupGetSubject();
+  private static final MethodHandle GET_CONTEXT = lookupGetContext();
 
-  private static MethodHandle lookupDoAs() {
+  private static MethodHandle lookupCallAs() {
     MethodHandles.Lookup lookup = MethodHandles.lookup();
     try {
       // Subject.doAs() is deprecated for removal and replaced by Subject.callAs().
@@ -133,7 +133,7 @@ public class SecurityUtils {
 
   /**
    * Get the current security manager, if available.
-   * @return the current security manager, if available
+   * @return the current security manager, if available, null otherwise
    */
   public static Object getSecurityManager() {
     try {
@@ -176,11 +176,10 @@ public class SecurityUtils {
    */
   public static <T> T doPrivileged(PrivilegedAction<T> action) {
     // Keep this method short and inlineable.
-    MethodHandle methodHandle = DO_PRIVILEGED;
-    if (methodHandle == null) {
+    if (DO_PRIVILEGED == null) {
       return action.run();
     }
-    return doPrivileged(methodHandle, action);
+    return doPrivileged(DO_PRIVILEGED, action);
   }
 
   @SuppressWarnings("unchecked")
@@ -190,6 +189,8 @@ public class SecurityUtils {
     } catch (RuntimeException | Error x) {
       throw x;
     } catch (Throwable x) {
+      // Unlikely to happen, as we catch Errors and RuntimeExceptions above, and doPrivileged
+      // does not throw checked exceptions
       throw new RuntimeException(x);
     }
   }
@@ -206,11 +207,11 @@ public class SecurityUtils {
   @SuppressWarnings("unchecked")
   public static <T> T callAs(Subject subject, Callable<T> action) {
     try {
-      MethodHandle methodHandle = DO_AS;
-      if (methodHandle == null) {
-        return action.call();
+      if (CALL_AS == null) {
+        throw new RuntimeException(
+            "Was unable to run either of Subject.callAs() or Subject.doAs()");
       }
-      return (T) methodHandle.invoke(subject, action);
+      return (T) CALL_AS.invoke(subject, action);
     } catch (RuntimeException | Error x) {
       throw x;
     } catch (Throwable x) {
@@ -234,7 +235,7 @@ public class SecurityUtils {
       MethodHandle methodHandle = CURRENT;
       return (Subject) methodHandle.invoke();
     } catch (Throwable x) {
-      throw new RuntimeException("Error when trying to get the current user", x);
+      throw new RuntimeException("Error while trying to get the current user", x);
     }
 
   }
@@ -243,9 +244,9 @@ public class SecurityUtils {
     return () -> {
       try {
         return callable.call();
-      } catch (RuntimeException x) {
+      } catch (RuntimeException | Error x) {
         throw x;
-      } catch (Exception x) {
+      } catch (Throwable x) {
         throw new RuntimeException(x);
       }
     };
@@ -253,11 +254,8 @@ public class SecurityUtils {
 
   private static Subject getSubjectFallback() {
     try {
-      MethodHandle contextMethodHandle = GET_CONTEXT;
-      Object context = contextMethodHandle.invoke();
-
-      MethodHandle getSubjectMethodHandle = GET_SUBJECT;
-      return (Subject) getSubjectMethodHandle.invoke(context);
+      Object context = GET_CONTEXT.invoke();
+      return (Subject) GET_SUBJECT.invoke(context);
     } catch (Throwable x) {
       throw new RuntimeException("Error trying to get the current Subject", x);
     }
--- a/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
@@ -166,8 +166,9 @@ public class SecurityUtils {
    * @param action the action to run
    * @return the result of the action
    * @param <T> the type of the result
+   * @throws CompletionException
    */
-  public static <T> T callAs(Subject subject, Callable<T> action) {
+  public static <T> T callAs(Subject subject, Callable<T> action) throws CompletionException {
     try {
       return (T) CALL_AS.invoke(subject, action);
     } catch (PrivilegedActionException e) {
--- a/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
@@ -179,7 +179,6 @@ public class SecurityUtils {
 
   /**
    * Maps to Subject.currect() is available, otherwise maps to Subject.getSubject()
-   * </p>
    * @return the current subject
    */
   public static Subject currentSubject() {
--- a/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
@@ -167,13 +167,11 @@ public class SecurityUtils {
    * @return the result of the action
    * @param <T> the type of the result
    */
-  public static <T> T callAs(Subject subject, Callable<T> action) throws Exception {
+  public static <T> T callAs(Subject subject, Callable<T> action) {
     try {
       return (T) CALL_AS.invoke(subject, action);
     } catch (PrivilegedActionException e) {
       throw new CompletionException(e.getCause());
-    } catch (Exception e) {
-      throw e;
     } catch (Throwable t) {
       throw sneakyThrow(t);
     }
--- a/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
@@ -167,11 +167,13 @@ public class SecurityUtils {
    * @return the result of the action
    * @param <T> the type of the result
    */
-  public static <T> T callAs(Subject subject, Callable<T> action) {
+  public static <T> T callAs(Subject subject, Callable<T> action) throws Exception {
     try {
       return (T) CALL_AS.invoke(subject, action);
     } catch (PrivilegedActionException e) {
       throw new CompletionException(e.getCause());
+    } catch (Exception e) {
+      throw e;
     } catch (Throwable t) {
       throw sneakyThrow(t);
     }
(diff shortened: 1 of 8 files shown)


stoty:
This is the alternate patch using reflection.

stoty:
Have you been able to check this @julianhyde  ?

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  @SuppressWarnings("unchecked")
+  private static <T> T doPrivileged(MethodHandle doPrivileged, PrivilegedAction<T> action) {
+    try {
+      return (T) doPrivileged.invoke(action);
+    } catch (RuntimeException | Error x) {
+      throw x;
+    } catch (Throwable x) {
+      throw new RuntimeException(x);
```
stoty, line 193: Note to self:
We should probably just fall back to wrapping and running, same as above.

stoty: No, because that would mask any real exception thrown.

stoty:
The approach here is directly lifted from Jetty (pruning some unused functions)

stoty:
rebased on main

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,266 @@
...
+   */
+  public static void checkPermission(Permission permission) throws SecurityException {
+    Object securityManager = SecurityUtils.getSecurityManager();
+    if (securityManager == null) {
+      return;
+    }
+    try {
+      securityManager.getClass().getMethod("checkPermission").invoke(securityManager,
+          permission);
+    } catch (SecurityException x) {
```
chrisdennis, line 162: I think this is wrong... it's being invoked via reflection so you'll need to catch `InvocationTargetException` and examine the contents.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,266 @@
...
+  public static <T> T callAs(Subject subject, Callable<T> action) {
+    try {
+      if (CALL_AS == null) {
+        throw new RuntimeException(
+            "Was unable to run either of Subject.callAs() or Subject.doAs()");
+      }
+      return (T) CALL_AS.invoke(subject, action);
+    } catch (RuntimeException | Error x) {
+      throw x;
+    } catch (Throwable x) {
```
chrisdennis, line 217: The `CompletionException` is thrown directly... why do we rewrap it?

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,266 @@
...
+    try {
+      MethodHandle methodHandle = CURRENT;
+      return (Subject) methodHandle.invoke();
+    } catch (Throwable x) {
+      throw new RuntimeException("Error while trying to get the current user", x);
+    }
+
+  }
+
+  private static <T> PrivilegedAction<T> callableToPrivilegedAction(Callable<T> callable) {
```
chrisdennis, line 243: Not investigated the wider context here... but I do wonder if this should map to `PrivilegedExceptionAction`?

chrisdennis (commented):
Since community review was requested I figured I'd cast an eye over this. The usage of `SecurityUtils` looks fine to me... the class itself has some issues for me though. Not sure how you want to approach that though, and obviously feel free to dismiss all of this if desired.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
stoty: Yes, this is a straight bug.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
stoty: The current code does not PrivilegedExceptionAction, so we don't need to implement that functionality.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
stoty: Should be fixed by unwrapInvocationTargetException.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
stoty: Should be fixed by unwrapInvocationTargetException.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,266 @@
...
+   * </p>
+   * @param subject the subject this action runs as
+   * @param action the action to run
+   * @return the result of the action
+   * @param <T> the type of the result
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> T callAs(Subject subject, Callable<T> action) {
+    try {
+      if (CALL_AS == null) {
```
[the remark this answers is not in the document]
stoty: That's a good point.

Generally, failing early is a good idea.

On the other hand, while I don't know any circumstance  where this error can happen (save perhaps a corrupted JVM) , not all use cases call these functions. For example, the client only ever needs SecurityUtils if Kerberos is used, so failing late could be seen as defensive programming.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,266 @@
...
+  private static MethodHandle lookupCallAs() {
+    MethodHandles.Lookup lookup = MethodHandles.lookup();
+    try {
+      // Subject.doAs() is deprecated for removal and replaced by Subject.callAs().
+      // Lookup first the new API, since for Java versions where both exists, the
+      // new API delegates to the old API (for example Java 18, 19 and 20).
+      // Otherwise (Java 17), lookup the old API.
+      return lookup.findStatic(Subject.class, "callAs",
+        MethodType.methodType(Object.class, Subject.class, Callable.class));
+    } catch (Throwable x) {
```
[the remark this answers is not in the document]
stoty: This is another gray area.
The current code has a higher chance of working in unexpected sitations.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,286 @@
...
+   */
+  public static <T> T callAs(Subject subject, Callable<T> action) {
+    try {
+      if (CALL_AS == null) {
+        throw new RuntimeException(
+            "Was unable to run either of Subject.callAs() or Subject.doAs()");
+      }
+      return (T) CALL_AS.invoke(subject, action);
+    } catch (InvocationTargetException x) {
+      return unwrapInvocationTargetException(x);
```
zabetak, line 215: The `invoke` methods on `MethodHandle` do not declare to throw an `InvocationTargetException`. Why do we choose to handle this case specifically? Are there some kind of best practices that we should follow?

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,286 @@
...
+
+  }
+
+  @SuppressWarnings("unused")
+  private static <T> PrivilegedAction<T> callableToPrivilegedAction(Callable<T> callable) {
+    return () -> {
+      try {
+        return callable.call();
+      } catch (InvocationTargetException x) {
+        return unwrapInvocationTargetException(x);
```
zabetak, line 249: Why should we handle the `InvocationTargetException` here?

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,286 @@
...
+    MethodHandles.Lookup lookup = MethodHandles.lookup();
+    try {
+      // Subject.doAs() is deprecated for removal and replaced by Subject.callAs().
+      // Lookup first the new API, since for Java versions where both exist, the
+      // new API delegates to the old API (for example Java 18, 19 and 20).
+      // Otherwise (Java 17), lookup the old API.
+      return lookup.findStatic(Subject.class, "callAs",
+        MethodType.methodType(Object.class, Subject.class, Callable.class));
+    } catch (Throwable x) {
+      try {
```
zabetak, line 52: nit: We could probably avoid the extra nesting here.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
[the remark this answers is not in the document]
chrisdennis: The current code will swallow things like "OutOfMemoryError", to me thats a real downside... but it's obviously not my call.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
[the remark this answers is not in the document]
chrisdennis: The SecurityUtil class isn't going to get loaded, and the static initialization won't run unless you get close enough to this code. So to me the window of "we'll load the SecurityUtil class but not use its functions" is not big enough to offset the downsides.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
chrisdennis: What I mean here is Callable<?> throws checked exceptions but PrivilegedAction doesn't. In that sense:

_`PrivilegedAction<T>` is to `Supplier<T>` as `PrivilegedExceptionAction<T>` is to `Callable<T>`_

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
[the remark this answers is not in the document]
zabetak: I guess we should at least put some logging in place wherever we shallow the exceptions.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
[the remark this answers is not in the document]
stoty: OK. I will change

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
stoty: unwrapInvocationTargetException is not needed, I will remove it.

chrisdennis:
I realized my suggestions for `SecurityUtils` were getting a little involved so I coded up the result and PRed here: https://github.com/stoty/calcite-avatica/pull/1

stoty:
I have applied your PR, and made some minor changes. These are in different commits.
Please check them @chrisdennis .

chrisdennis (approved):
Looks good to me... you rightly point out that what we have makes some assumptions about the order of any future removals but I think they're reasonable assumptions.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,286 @@
...
+   * @param action the action to run
+   * @return the result of the action
+   * @param <T> the type of the result
+   */
+  public static <T> T callAs(Subject subject, Callable<T> action) {
+    try {
+      if (CALL_AS == null) {
+        throw new RuntimeException(
+            "Was unable to run either of Subject.callAs() or Subject.doAs()");
+      }
```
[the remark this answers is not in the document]
stoty: This is no longer the case.

server/src/test/java/org/apache/calcite/avatica/SpnegoTestUtil.java
```diff
@@ -138,8 +138,9 @@ public static void refreshJaasConfiguration() {
...
     // has been initialized. We need to nuke that static instance to make sure our
     // serverSpnegoConfigFile gets read.
-    AccessController.doPrivileged(new PrivilegedAction<Configuration>() {
-      public Configuration run() {
+    SecurityUtils.doPrivileged(new PrivilegedAction<Configuration>() {
+      @Override
+    public Configuration run() {
         return Configuration.getConfiguration();
       }
     }).refresh();
```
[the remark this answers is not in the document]
stoty: Unhidden by accident

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,202 @@
...
+  @SuppressWarnings("unused")
+  private static <T> PrivilegedExceptionAction<T> callableToPrivilegedExceptionAction(
+      Callable<T> callable) {
+    return callable::call;
+  }
+
+  @SuppressWarnings("unchecked")
+  private static <E extends Throwable> RuntimeException inferAndCast(Throwable e) throws E {
+    throw (E) e;
+  }
```
zabetak, line 201: The current name is misleading since the method does not actually infer something; it does not really cast either. 
This pattern that allows to throw a checked exception without declaring in the method is usually known as "sneakyThrow". I am not sure how many people are familiar with this hack so it may be better to rename this method accordingly.

I think its also possible to change the return type to `void`  and possibly change `throw inferAndCast(t)` to `sneakyThrow(t)`.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,266 @@
...
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.security.Permission;
+import java.security.PrivilegedAction;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletionException;
+import javax.security.auth.Subject;
+
+/**
+ * This class is heavily based on SecurityUtils in Jetty 12.0
```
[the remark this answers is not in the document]
zabetak: The provided links are not about source headers and file modifications but more about how to document which license is picked when more than one apply to the work.

Having said the class already differs quite a bit from the original one present in Jetty (after the multiple review rounds) so it may no longer be that pertinent to consider it as a copy/third-party work.

Anyways since the original Jetty code is also under AL2 this discussion may not be that important after all.

zabetak (commented):
Few small comments and I think we are ready to merge this.

core/src/main/java/org/apache/calcite/avatica/util/SecurityUtils.java
```diff
@@ -0,0 +1,202 @@
...
+      Class<?> klass =
+          ClassLoader.getSystemClassLoader().loadClass("java.security.AccessController");
+      MethodHandles.Lookup lookup = MethodHandles.lookup();
+      return lookup.findStatic(klass, "doPrivileged",
+        MethodType.methodType(Object.class, PrivilegedAction.class));
+    } catch (NoSuchMethodException | IllegalAccessException x) {
+      // Assume that single methods won't be removed from AcessController
+      throw new AssertionError(x);
+    } catch (ClassNotFoundException e) {
+      return null;
```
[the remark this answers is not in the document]
chrisdennis: The calling code cannot handle `AccessController.getContext()` being missing, but it can handle `AccessController.doPrivileged()` being missing (it falls back to a straight call). I can think of two ways to clarify this (if it's deemed necessary):

1. Clarify the method names: `tryLookupDoPrivileged()` vs `lookupGetContext()` or maybe `lookupGetContextOrFail()`.
2. Return optional from all the lookup methods and move the "failure behavior" to the caller, where it is clearer what is happening, (albeit doing this without losing the failure context/information is a little tricky).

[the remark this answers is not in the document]
zabetak: If the `AccessController` class is missing then initialization of this class will fail so not sure why we need a distinction.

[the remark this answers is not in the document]
zabetak: OK then that's fine with me we can leave the code as is. Thanks for the clarifications!

stoty:
I have changed the name back to sneakyThrow().

Since we're going for perfection, we may also want to reconsider callAs().

Subject callAs() throws Exceptions, and the fallback path now also throws exceptions (wrapped in PrivilegedActionException).

We could change it to something like :

  public static <T> T callAs(Subject subject, Callable<T> action) throws Exception {
    try {
      return (T) CALL_AS.invoke(subject, action);
    } catch (PrivilegedActionException e) {
      e.getException();
    } catch (Exception e) {
      throw e;
    } catch (Throwable t) {
      throw sneakyThrow(t);
    }
  }

chrisdennis:
> Since we're going for perfection, we may also want to reconsider callAs().

Apologies for triggering some (all?) of this bike-shedding.

> Subject callAs() throws Exceptions, and the fallback path now also throws exceptions (wrapped in PrivilegedActionException).
> 

I'm confused, Subject.callAs(...) throws `CompletionException` if the action throws, which is what the current code does, unwrapping any PrivilegedActionException and rewrapping it in case we're backed by the old `doAs(...)` method. The only tweak I'd make to the current code is that the `Exception` catch and rethrow is redundant given the `sneakyThrow(t)` call.

stoty:
> I'm confused, Subject.callAs(...) throws `CompletionException` if the action throws, which is what the current code does, unwrapping any PrivilegedActionException and rewrapping it in case we're backed by the old `doAs(...)` method. The only tweak I'd make to the current code is that the `Exception` catch and rethrow is redundant given the `sneakyThrow(t)` call.

You are right.

However, since Subject.callAs() explicitly throws CompletionException, shouldn't SecurityUtils.callAs() also explictly throw CompletionException ?

chrisdennis:
> However, since Subject.callAs() explicitly throws CompletionException, shouldn't SecurityUtils.callAs() also explictly throw CompletionException ?

Yes, that would make sense.

stoty:
This is as polished as it gets now. Can you do a final review @zabetak  ?

zabetak (approved):
The changes in the PR LGTM!

I also run `./gradlew build` with JDK23 locally and it passed without problems so I guess we are good to merge this.

stoty:
Merged manually.
(Maybe we could enable squash merges...)

F21:
Thanks for getting this in @stoty, @chrisdennis and @zabetak!

PR #257, opened 2024-10-31
[CALCITE-6658] Update Jackson from 2.15.2 to 2.15.4 in Avatica

stoty:


stoty:
This did not fail for me when testing the upgrades together.
Possibly the checkstyle update pulls in a new enough asm jar ?

stoty:
Rebasing on top of #260 only does not help.
Is some other plugin forcing an asm downgrade ?

stoty:
The problem is that the current asm version constraints apply to the code buing built, but they do not apply to the plugins itself.

Asm is only used in tests by kerby, and it doesn't seem to cause a problem, it may even be necessary, so I wouldn't touch that, but we definitely need to add the constraint for the shadow plugin.

No why it worked for me when I first tested it, probably I made some mistake.

stoty:
Technically, we have two mechanisms:
- The constraints for the project ( not build), are applied by the property reading magic.
- We have a block for applying the constraints for the spotbug plugins only.

There is nothing that that would apply to the shadowJar plugin.

The mechanism used for overriding the constraints does not work for shadowJar plugin, but we can just add the constraint globall to the build environment.

zabetak (approved):
+1 Pending tests

PR #264, opened 2024-11-12
[CALCITE-6687] Add dependency constraints for org.ow2.asm for Gradle …

stoty:
…build environment

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -31,6 +31,19 @@ import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApis
 import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApisExtension
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 
+buildscript {
+    dependencies {
+        constraints {
+            "classpath"("org.ow2.asm:asm:${property("asm.version")}")
+            "classpath"("org.ow2.asm:asm-all:${property("asm.version")}")
+            "classpath"("org.ow2.asm:asm-analysis:${property("asm.version")}")
+            "classpath"("org.ow2.asm:asm-commons:${property("asm.version")}")
+            "classpath"("org.ow2.asm:asm-tree:${property("asm.version")}")
+            "classpath"("org.ow2.asm:asm-util:${property("asm.version")}")
+        }
+    }
+}
+
 plugins {
     publishing
     // Verification
@@ -344,16 +357,6 @@ allprojects {
                 excludeFilter = file("$rootDir/src/main/config/spotbugs/spotbugs-filter.xml")
                 // By default spotbugs verifies TEST classes as well, and we do not want that
                 this.sourceSets = listOf(sourceSets["main"])
-            }
-            dependencies {
-                constraints {
-                    "spotbugs"("org.ow2.asm:asm:${"asm".v}")
-                    "spotbugs"("org.ow2.asm:asm-all:${"asm".v}")
-                    "spotbugs"("org.ow2.asm:asm-analysis:${"asm".v}")
-                    "spotbugs"("org.ow2.asm:asm-commons:${"asm".v}")
-                    "spotbugs"("org.ow2.asm:asm-tree:${"asm".v}")
-                    "spotbugs"("org.ow2.asm:asm-util:${"asm".v}")
-                }
             }
         }
 


build.gradle.kts
```diff
@@ -31,6 +31,21 @@ import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApis
 import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApisExtension
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 
+buildscript {
+
+    dependencies {
+        constraints {
+            // FIXME Should use asm.version from properties
+            "classpath"("org.ow2.asm:asm::9.7.1")
```
F21, line 39: Not too familiar with gradle, but is there anyway to avoid hardcoding the versions in here?

stoty: Neither I am.
I have tried a few things, but I was not able to get anything working.

stoty: So if there is one, I couldn't find it.

stoty: I suspect that as this is used for initializing gradle, you cannot even access properties or run kotlin code in this block.

Maybe I should remove the FIXME comment, as using properties here may not even be possible.

build.gradle.kts
```diff
@@ -31,6 +31,20 @@ import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApis
...
+            "classpath"("org.ow2.asm:asm:9.7.1")
+            "classpath"("org.ow2.asm:asm-all:9.7.1")
+            "classpath"("org.ow2.asm:asm-analysis:9.7.1")
+            "classpath"("org.ow2.asm:asm-commons:9.7.1")
+            "classpath"("org.ow2.asm:asm-tree:9.7.1")
+            "classpath"("org.ow2.asm:asm-util:9.7.1")
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
stoty: Unfortunately, those are not plugin versions.
These are overrides for the dependencies of a plugin.

And we cannot use a newer plugin, because this is the last version of the shade plugin than runs on Java 8.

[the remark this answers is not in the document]
stoty: So we must use an older plugin, but we need newer asm for it to be able to process newer class files.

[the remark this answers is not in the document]
zabetak: I was under the impression that you can define constraints inside `pluginManagement` but after doing some research it turns out that its not possible.

Basically, I reached out to the same conclusion as you. The only way to select dependencies for plugins is via the `buildscript` block.

build.gradle.kts
```diff
@@ -31,6 +31,20 @@ import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApis
 import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApisExtension
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 
+buildscript {
+    dependencies {
+        constraints {
+            // Looks like we cannot access the versions defined in gradle.properties from here
```
zabetak, line 37: We can access the properties in `gradle.properties` if we use the built-in functions. For instance the following works:
```
"classpath"("org.ow2.asm:asm:${property("asm.version")}")
```

gradle.properties
```diff
@@ -55,6 +55,8 @@ checkstyle.version=10.19.0
 jdk8.checkstyle.version=9.3
 spotbugs.version=3.1.11
 
+# This only applies to the asm used in the tests. To override the asm used by the gradle plugins
+# change the version in the buildScript block of build.gradle.kts directly
```
zabetak, line 59: After using the `${property("asm.version")}` syntax this comment is no longer necessary.

zabetak (changes requested):
LGTM, the only thing that I would change before merging is to read the property from the `gradle.properties` file as suggested.

On a more general topic, we should not be forced to fine-tune deps for plugins but rather upgrade the version plugin itself. I understand that we cannot do it now since JDK8 is unsupported in more recent version but I guess we should start discussing till when we are going to support the JDK8 on our side. Eventually we are gonna reach a point that workarounds like this one are not gonna work.

stoty:
> How we can be sure that constraints are enforced correctly? In other words, how do we test this PR?

#257 fails with the default asm version. If that build successfully, then the constraints are enforced.

zabetak:
We can also check that constraints are enforced via `./gradlew buildEnvironment` or by publishing a build scan.

stoty:
Thank you, I have added the suggested property references, this is a much better solution.

As for JDK 8 support, I agree that it is generally a pain, and is only getting worse, but some projects using Avatica still support JDK8.

zabetak (changes requested):

zabetak (approved):

stoty:
rebased and merged commits

PR #265, opened 2024-11-17
[CALCITE-6697] Update Scott-data-hsqldb from 0.1 to 0.2 in Avatica

caicancai:
https://issues.apache.org/jira/browse/CALCITE-6697

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -76,6 +76,6 @@ kerby.version=1.1.1
 log4j2.version=2.17.1
 mockito.version=4.11.0
 protobuf.version=3.25.5
-scott-data-hsqldb.version=0.1
+scott-data-hsqldb.version=0.2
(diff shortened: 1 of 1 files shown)


stoty:
The test failures look related.
Maybe hsqldb.version also needs to be synced ?

caicancai:
> The test failures look related. Maybe hsqldb.version also needs to be synced ?

I'll try it later, thanks for the reminder

caicancai:
@stoty It looks like some of the data in the test set has been transformed. I’m not sure whether to continue this work. Do you have any suggestions?

stoty:
Is the hsql update a security issue ?
If not, I'd just leave it alone, it's only used for tests.

Otherwise, all the tests using the dataset and hsql would need to be checked and updated to work with the new HSQL and test data.

PR #268, opened 2025-01-23
[CALCITE-6590] Handle security deprecation using code templates

zabetak:


zabetak:
Testing an alternative approach to the reflection based in #251

stoty:
A possible problem with this approach is that I seem to recall that the shade plugin breaks code templates (i.e. they don't work in shaded jars)

zabetak:
I wrote "template" but its not a correct term. It's basically JDK specific java files; one will be included when compiling with JDK >= 18 and the other when JDK < 18 so I don't see why they shouldn't work.

stoty:
Right, I was thinking of https://openjdk.org/jeps/238 , and https://issues.apache.org/jira/browse/MSHADE-406
(though MSHASE-406 MAY be fixed now ?)

A compile-time solution would only make sense if we shipped separate artifacts compiled for JDK18, and would accept that the non-JDK18 artifact won't work on JDK23+.

I think the goal is to have single JAR that works on JDK8-JDK23+.

JEP238 would be a solution if the shade plugin can handle it,  but I never tested that.
I expect that it would also complicate the build a great deal.
(We should also check gradle shadow plugin if we want to go there)

zabetak:
> I think the goal is to have single JAR that works on JDK8-JDK23+.
If that's the case then indeed the approach here will not work. We should maybe clarify this part with the rest of the community.

At the moment, the CI does not test cross-JDK compilation/execution so I am not sure if we should claim that the JDK8 compiled jars are OK to run with JDK23+. All the CI checks are using the same JDK for compiling the project and running the tests.

stoty:
Having a single binary that works with all JVMs is the only solution that doesn't require extra steps from users.
For example, we want users to be able to update their JVM from 17 to 23, without having to replace JARs, and want to avoid having JVM specific distributions both for Avatica, and its dependencies.

This is also the approach taken by projects that I am aware of.

As for testing, running the test suite cross-JVM would be nice, especially as the suite is not expensive.
I don't think it's critical though, as every code path uses reflectcion, so the existing tests already give us full coverage.

We should definitely bump the JDK22 CI run to use JDK23; either in the JDK23 support patch, or in separate JIRA.

zabetak:
I guess the reflection approach is more favorable thus I am closing this for now.

PR #272, opened 2025-01-28
[CALCITE-5136] Avatica build (or CI) must fail if there are deprecation warnings

chrisdennis:
Fixes applied are:
 * move `new URL(...)` to `new URI(...).toUrl()`
 * move `Assert.assertThat(...)` to `MatcherAssert.assertThat(...)`
 * move `ExpectedException` ~to `Assert.assertThrows(...)`~ use manual code since Guava version changes cause the code to see both 4.12 and 4.13.1 JUnit versions.
 * in all other cases either propagate the deprecation or suppress it.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/util/Sources.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/Sources.java
@@ -59,8 +59,8 @@ public abstract class Sources {
 
   public static Source url(String url) {
     try {
-      return of(new URL(url));
-    } catch (MalformedURLException | IllegalArgumentException e) {
+      return of(new URI(url).toURL());
+    } catch (MalformedURLException | IllegalArgumentException | URISyntaxException e) {
       throw new RuntimeException("Malformed URL: '" + url + "'", e);
     }
   }
@@ -133,11 +133,7 @@ public abstract class Sources {
           filePath += "/";
         }
         try {
-          // We need to encode path. For instance, " " should become "%20"
-          // That is why java.net.URLEncoder.encode(java.lang.String, java.lang.String) is not
-          // suitable because it replaces " " with "+".
-          String encodedPath = new URI(null, null, filePath, null).getRawPath();
-          return new URL("file", null, 0, encodedPath);
+          return new URI("file", filePath, null).toURL();
         } catch (MalformedURLException | URISyntaxException e) {
           throw new IllegalArgumentException("Unable to create URL for file " + filePath, e);
         }
--- a/server/src/test/java/org/apache/calcite/avatica/server/HttpServerSpnegoWithoutJaasTest.java
+++ b/server/src/test/java/org/apache/calcite/avatica/server/HttpServerSpnegoWithoutJaasTest.java
@@ -45,7 +45,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
-import java.net.URL;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.security.Principal;
 import java.util.Properties;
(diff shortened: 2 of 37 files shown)


stoty:
> move ExpectedException to Assert.assertThrows(...) use manual code since Guava version changes cause the code to see both 4.12 and 4.13.1 JUnit versions.

Are you sure ?  We're dependency managing Junit to 4.12 , and I could not find 4.13.1 in the gradle dependecy output anywhere ?

We should upgrade to  4.13.2 anyway (in a different JIRA)

server/src/test/java/org/apache/calcite/avatica/server/HttpServerSpnegoWithoutJaasTest.java
```diff
@@ -81,7 +81,7 @@ public class HttpServerSpnegoWithoutJaasTest {
   private static boolean isKdcStarted = false;
   private static boolean isHttpServerStarted = false;
 
-  private static URL httpServerUrl;
```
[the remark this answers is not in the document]
chrisdennis: Pushed the rename. I *think* this was the only case where we changed the type.

chrisdennis:
FYI: I pounded on the test that failed on macOS locally and see the following pattern:
```
Failure Statistics
49976 : connection refused
24 : connection reset
```
The existing code doesn't account for "connection reset". I have a fix for this... does it need a JIRA issue, and a separate PR?

stoty:
Thank you.

Preferably yes. CALCITE-6799 already exists for the issue.

core/src/main/java/org/apache/calcite/avatica/util/Sources.java
```diff
@@ -136,8 +136,7 @@ private static URL fileToUrl(File file) {
           // We need to encode path. For instance, " " should become "%20"
```
stoty, line 136: nit:
Is this comment still needed ?

stoty (approved):
+1 LGTM

core/src/main/java/org/apache/calcite/avatica/util/Sources.java
chrisdennis: No, probably not. Removed... but closer inspection here forced me to fix something else. This branch of the code is making a very "odd" URI/URL. It's hierarchical URI, with a relative and a scheme. I don't know if they're illegal under the RFC... but it's certainly weird. I had to subtly tweak things here to construct one by treating it as a non-heirarchical URL (args here are: scheme, scheme-specific-part, fragment). The behavior now matches the old code.

rubenada:
This PR looks in a good shape, shall we move on?
One minor detail though, per our contribution [guidelines](https://calcite.apache.org/develop/#contributing), commit messages should not end with a period; apart from that I think generally we tend to use as final commit message the Jira title, i.e. `[CALCITE-5136] Avatica build (or CI) must fail if there are deprecation warnings` in this case.

chrisdennis:
> One minor detail though, per our contribution [guidelines](https://calcite.apache.org/develop/#contributing), commit messages should not end with a period; apart from that I think generally we tend to use as final commit message the Jira title, i.e. `[CALCITE-5136] Avatica build (or CI) must fail if there are deprecation warnings` in this case.

Fixed.

chrisdennis:
@stoty something is up in CI... the force push didn't change anything: https://github.com/apache/calcite-avatica/compare/da93125c4a177717611beb1db0b51f3bf279e390..fd947c724f8a46ec45f5126adcefdeb4868e25eb 

Yet somehow everything is now failing... and it looks like it's because the build resolved 4.13.x of JUnit (evidenced in all the `assertThat(...)` deprecation warnings).

.idea/vcs.xml
```diff
@@ -1,27 +1,27 @@
...
-                    <option name="linkRegexp" value="https://github.com/apache/calcite-avatica/pull/$2" />
-                </IssueNavigationLink>
-            </list>
-        </option>
-    </component>
-    <component name="VcsDirectoryMappings">
-        <mapping directory="$PROJECT_DIR$" vcs="Git" />
-    </component>
-</project>
+  <component name="CommitMessageInspectionProfile">
```
stoty, line 3: Did you mean to include this @chrisdennis  ?

It doesn't seem related to the depracations.

chrisdennis: I did not... I guess IntelliJ got over-eager with the reformatting and I did an ill-advised `git add .`.

stoty:
merged manually.

PR #275, opened 2025-02-02
[CALCITE-5094] Calcite JDBC Adapter and Avatica should support MySQL UNSIGNED types of TINYINT, SMALLINT, INT, BIGINT

strongduanmu:
For more details, refer jira: https://issues.apache.org/jira/browse/CALCITE-5094

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/AvaticaSite.java
+++ b/core/src/main/java/org/apache/calcite/avatica/AvaticaSite.java
@@ -291,7 +291,7 @@ public class AvaticaSite {
 
   /** Similar logic to {@link #setObject}. */
   public static Object get(Cursor.Accessor accessor, int targetSqlType,
-      Calendar localCalendar) throws SQLException {
+      boolean signed, Calendar localCalendar) throws SQLException {
     switch (targetSqlType) {
     case Types.CLOB:
     case Types.DATALINK:
@@ -304,6 +304,9 @@ public class AvaticaSite {
     case Types.ARRAY:
       return accessor.getArray();
     case Types.BIGINT:
+      if (!signed) {
+        return accessor.wasNull() ? null : accessor.getULong();
+      }
       final long aLong = accessor.getLong();
       if (aLong == 0 && accessor.wasNull()) {
         return null;
@@ -337,6 +340,9 @@ public class AvaticaSite {
       }
       return aDouble;
     case Types.INTEGER:
+      if (!signed) {
+        return accessor.wasNull() ? null : accessor.getUInt();
+      }
       final int anInt = accessor.getInt();
       if (anInt == 0 && accessor.wasNull()) {
         return null;
@@ -362,6 +368,9 @@ public class AvaticaSite {
     case Types.ROWID:
       throw notImplemented();
     case Types.SMALLINT:
+      if (!signed) {
+        return accessor.wasNull() ? null : accessor.getUShort();
+      }
       final short aShort = accessor.getShort();
       if (aShort == 0 && accessor.wasNull()) {
         return null;
@@ -372,6 +381,9 @@ public class AvaticaSite {
     case Types.TIMESTAMP:
       return accessor.getTimestamp(localCalendar);
     case Types.TINYINT:
+      if (!signed) {
+        return accessor.wasNull() ? null : accessor.getUByte();
+      }
       final byte aByte = accessor.getByte();
       if (aByte == 0 && accessor.wasNull()) {
         return null;
--- a/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
+++ b/core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
@@ -20,6 +20,11 @@ import org.apache.calcite.avatica.AvaticaSite;
 import org.apache.calcite.avatica.AvaticaUtils;
 import org.apache.calcite.avatica.ColumnMetaData;
 
+import org.joou.UByte;
+import org.joou.UInteger;
+import org.joou.ULong;
+import org.joou.UShort;
+
 import java.io.InputStream;
 import java.io.Reader;
 import java.lang.reflect.Field;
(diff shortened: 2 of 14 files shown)


mihaibudiu (approved):
The changes look fine, but it's hard to tell whether all code paths that may be relevant have been handled until we have some end-to-end tests.

strongduanmu:
> The changes look fine, but it's hard to tell whether all code paths that may be relevant have been handled until we have some end-to-end tests.

Yes, I agree with you, end-to-end testing is very important. Currently, I have tested different unsigned types with `calcite avatica 1.26.0-SNAPSHOT` and they all work fine. When this PR is merged, I will try to support unsigned types in calcite and add more end-to-end test cases.

mihaibudiu:
I am wondering whether it will pay off to actually build a set of classes UnsignedInt, UnsignedShort, etc, which would encapsulate these abstractions in a clean way. I am not sure that this is the right place for them - BTW. But I think you will need them eventually in the runtime of Calcite to properly implement the semantics of unsigned arithmetic.

mihaibudiu:
I notice that Guava has some of these.

strongduanmu:
Hi @mihaibudiu, I spent some time looking at Guava. Currently, it only provides `UnsignedInteger` and `UnsignedLong`. Other `UnsignedByte` and `UnsignedShort` are not provided by Guava. 

In addition, I found that MySQL provides another idea, we can handle them by mapping a wider range of Java types. For example, the `BIGINT` type is usually mapped to `java.lang.Long` for processing, while `BIGINT UNSIGNED` needs to be mapped to `java.math.BigInteger` for processing. For more type mappings, please refer to the document: https://dev.mysql.com/doc/connector-j/en/connector-j-reference-type-conversions.html

Through this mapping conversion method, we do not need to add any classes, and the Calcite runtime can handle UNSIGNED types. What do you think of this approach? I will keep trying if you have any suggestions.

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
```diff
@@ -145,28 +145,46 @@ private boolean getColumn(final ResultSet resultSet, final int index,
       case 3: // SMALLINT
         resultSet.getShort(index);
         break;
-      case 4: // INTEGER
+      case 4: // MEDIUMINT
         resultSet.getInt(index);
         break;
-      case 5: // BIGINT
+      case 5: // INTEGER
```
ILuffZhe, line 151: case 4 and case 5 can be merged

core/src/main/java/org/apache/calcite/avatica/AvaticaSite.java
```diff
@@ -291,7 +291,7 @@ public void setObject(Object x, int targetSqlType) {
 
   /** Similar logic to {@link #setObject}. */
   public static Object get(Cursor.Accessor accessor, int targetSqlType,
-      Calendar localCalendar) throws SQLException {
+                           boolean signed, Calendar localCalendar) throws SQLException {
```
ILuffZhe, line 294: Nit: the indent, so are the following changes below in MetaImpl.

ILuffZhe (approved):
Overall looks good.

core/src/test/java/org/apache/calcite/avatica/AvaticaResultSetThrowsSqlExceptionTest.java
strongduanmu: Great suggestion, done.

core/src/main/java/org/apache/calcite/avatica/AvaticaSite.java
strongduanmu: Thank you, I have modified the indent.

mihaibudiu:
From my experience, in the long term it pays off to implement the right abstractions.
I expect there will be a lot of work to be done using unsigned types, including in the type checker, expression simplification, and runtime system. In particular, at runtime the enumerable implementation looks at the runtime types to make some decisions (maybe that's a bug in the implementation). So if the runtime will see a Long it won't know whether it's a real Long or an UnsignedInt in disguise.

strongduanmu:
> From my experience, in the long term it pays off to implement the right abstractions. I expect there will be a lot of work to be done using unsigned types, including in the type checker, expression simplification, and runtime system. In particular, at runtime the enumerable implementation looks at the runtime types to make some decisions (maybe that's a bug in the implementation). So if the runtime will see a Long it won't know whether it's a real Long or an UnsignedInt in disguise.

Hi @mihaibudiu, thank you for sharing these valuable experiences. These are issues I had not considered before. Since I am still a novice, I will spend some time to understand type checking, expression simplification, etc., and then try to add new UnsignedInt, UnsignedLong and other types.

strongduanmu:
> Can we merge this PR? Frankly, we won't know whether this stuff works until there are some end-to-end tests, so it would be nice to have a confirmation that Calcite will be able to use this.

Hi @mihaibudiu, according to your review suggestion, we'd better introduce unsigned type to facilitate Calcite to do some boundary checking on types. This part of work is not included in the current PR. I suggest merging this PR first, and then I will introduce https://github.com/jOOQ/jOOU in the next PR to support unsigned type. What do you think?

mihaibudiu:
I don't know if Avatica needs to take a dependency on joou.
Can you write some tests in Calcite that confirm that this PR works as expected? 
Such tests can be merged after this PR is merged; you will have to run them by linking Calcite with an unreleased version of Avatica.

strongduanmu:
> There is now [apache/calcite#4411](https://github.com/apache/calcite/pull/4411); I guess the tests should use both PRs.

Yes, I will test in calcite when #4411 is merged.

core/src/main/java/org/apache/calcite/avatica/util/AbstractCursor.java
```diff
@@ -288,18 +293,38 @@ public byte getByte() throws SQLException {
...
+      return UByte.valueOf(getLong());
+    }
+
     public short getShort() throws SQLException {
       return (short) getLong();
     }
 
+    @Override
+    public UShort getUShort() throws SQLException {
+      return UShort.valueOf(String.valueOf(getLong()));
```
mihaibudiu, line 307: this seems a bit convoluted

core/src/test/java/org/apache/calcite/avatica/CursorFactoryDeduceTest.java
```diff
@@ -244,6 +260,88 @@ protected static class SimplePOJO {
...
     }
   }
+
+  @Test public void deduceRecordCursorFactoryProjectedSignedField() throws Exception {
+    List<ColumnMetaData> columnsMetaDataList = Arrays.asList(
+        MetaImpl.columnMetaData("byteField", 1, ColumnMetaData.scalar(
+            Types.TINYINT, "TINYINT", ColumnMetaData.Rep.BYTE), true, true),
+        MetaImpl.columnMetaData("shortField", 2, ColumnMetaData.scalar(
+            Types.SMALLINT, "SMALLINT", ColumnMetaData.Rep.SHORT), true, true),
+        MetaImpl.columnMetaData("intField", 3, ColumnMetaData.scalar(
```
mihaibudiu, line 270: this we don't have, right?

mihaibudiu (approved):

mihaibudiu:
Please rebase and squash the commits.

PR #278, opened 2025-02-12
[CALCITE-6828] Upgrade to Kerby from 1.1.1 to 2.1.0 in Avatica

stoty:


stoty:
The Calcite failure is caused by the same maven problem that this fixes in Avatica.

rubenada (approved):

stoty:
Megred manually with fixed commit message.

PR #280, opened 2025-02-25
[CALCITE-6855] Report elapsed time in ConnectionPropertiesHATest assert

stoty:


stoty:
Will track this under the main JIRA #281 .

PR #284, opened 2025-07-14
[CALCITE-1480] Support specifying ciphersuites and algorithms for TLS

stoty:


joshelser (approved):
pretty far away from avatica these days, but this looks reasonable.

F21 (approved):

stoty:
fixed author

PR #285, opened 2025-07-23
CALCITE-3533 - Map non-jdbc data types to ANY in JdbcSchema

jeremyosterhoudt:
This PR allows non-standards JDBC/SQL types to be processed as `ANY`.  Databases such as Oracle and SQLServer (among others) often support SQL types that are not defined as part of the standard Java JDBC type mapping.  

Currently Avatica (and by proxy Calcite) throw an exception when these types are encountered.  However, for many non-standard types this prevents the query from executing successfully even though the results are retrieved and displayed correctly when `ANY` is used for these scenarios (SQLSever's DataTimeOffset for example).  

While this is not a fix per se I do believe this is an improvement over the current behavior.  Types could be added directly, however, some types like Oracle's `DateTime With Offset` require the Oracle JDBC driver to convert.  This would mean including the JDBC driver in Avatica with is likely undesirable.  Ideally I think a nice enhancement would be to allow developers/users to specify a non-standard type conversion class/factory that could be called when these types are encountered while embedding any needed libraries (I'm hoping to have a conversation around this and partake in the implementation).

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/SqlType.java
+++ b/core/src/main/java/org/apache/calcite/avatica/SqlType.java
@@ -378,7 +378,11 @@ public enum SqlType {
   public static SqlType valueOf(int type) {
     final SqlType sqlType = BY_ID.get(type);
     if (sqlType == null) {
-      throw new IllegalArgumentException("Unknown SQL type " + type);
+      // In most cases the ANY type will allow non-standard types to be displayable in the query requests.
+      // However, I think we should add the ability to specify a non-standard type factory
+      // that can be referenced here to see if the type is supported.
+      // For now, we will just return ANY to unblock users of databases that support non-standard types.
+      return ANY;
     }
     return sqlType;
   }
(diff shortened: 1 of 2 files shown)


mihaibudiu (commented):
The PR title does not match the issue title.
Also, I find the title confusing: returned by whom?
Maybe you can update the JIRA issue to explain what this achieves.
Is there a test to show what new capabilities this offers?

mihaibudiu:
These unit tests are not particularly illuminating.
I see that avatica has a dependency on HSQLDB, is there a test you can write using this database that tests the new capability end-to-end?

rubenada:
@jeremyosterhoudt would it be possible to fix the check-style of the PR, and also have some end-to-end test showing the fix (as suggested by @mihaibudiu ) ?

zabetak (commented):
It seems that the easier way to showcase the problem (`IllegalArgumentException`) in a more realistic setting would be through Calcite. @jeremyosterhoudt In addition to the PR here can you come up with a PR in the calcite code base that adds relevant test(s) (possibly in https://github.com/apache/calcite/blob/main/core/src/test/java/org/apache/calcite/test/JdbcTest.java).

In some cases a failure/crash is preferred, compared to returning wrong results or causing data corruption so we should ensure that the fix here will not cause some other undesired side effects.

PR #287, opened 2025-09-09
[CALCITE-7165] Update OWASP plugin from 10.0.4 to 12.13 in Avatica

stoty:


zabetak:
In Calcite build scripts we have [some conditional parts](https://github.com/apache/calcite/blob/53d20a5420a5d4b53d7099ff6d1bf31a22946775/build.gradle.kts#L524) based on the JavaVersion so maybe you can take some inspiration from there for making the plugin conditional.

PR #292, opened 2025-09-09
[CALCITE-7170] Support ISO String Binding for Date/Time/Timestamp in Avatica

sanskritimbora:
I would like to propose adding support for binding ISO-formatted string values to date, time, and timestamp types in Avatica. Currently, Avatica only supports binding these types using Java objects (such as java.sql.Date, java.sql.Time, and java.sql.Timestamp), which can be limiting for clients that work with standard string representations.

Adding ISO string binding would improve interoperability and make it easier to work with standard date/time formats across different clients and languages.

 
Simple client use case example:
```java
try (var connection = DriverManager.getConnection("jdbc:arrow-flight://localhost:5000;useEncryption=false");
{{ PreparedStatement createSchemaStatement = connection.prepareStatement("create schema if not exists foo");}}
{{ PreparedStatement createTableStatement = connection.prepareStatement("create table if not exists foo.bar (t time, d date, ts timestamp)");}}
{{ PreparedStatement insertStatement = connection.prepareStatement("insert into test values (?, ?, ?)")}}
) {
{{ createSchemaStatement.execute();}}
{{ createTableStatement.execute();}}

{{ insertStatement.setString(1, "2025-08-14T15:53:00.000Z");}}
{{ insertStatement.setString(2, "2025-04-01");}}
{{ insertStatement.setString(3, "21:39:50");}}
{{ }}
{{ insertStatement.executeUpdate();}}
}
```

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/AvaticaSite.java
+++ b/core/src/main/java/org/apache/calcite/avatica/AvaticaSite.java
@@ -37,6 +37,8 @@ import java.sql.SQLXML;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.sql.Types;
+import java.time.*;
+import java.time.format.DateTimeParseException;
 import java.util.Calendar;
 
 /**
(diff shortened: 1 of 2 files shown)


stoty:
This is a super hairy topic.

One issue is that "2025-08-14T15:53:00.000Z" doe not map to java.sql.Types.TIMESTAMP , it maps to java.sql.Types.TIMESTAMP_WITH_TIMEZONE .

Please add End-to-End tests to make sure that the with/without timezone conversion rules are properly applied.

See https://medium.com/@stoty/a-bug-for-ages-fixing-time-zone-handling-in-apache-phoenix-e9934d7acd80 
for an in-depth discussion of the topic, and templates for the end to end tests.

PR #294, opened 2025-09-10
[CALCITE-7165] Upgrade OWASP plugin version to 12.1.3 for JDKs > 8

zabetak:


stoty:
Tested with Java 8, The OWASP plugin fails, but the build generally works, which is a good as we can get.
It works with Java 11.

stoty:
I messed up and commited this together with one of my changes.

Can you please open a follow-up ticket to handle Java 8 and 9 @zabetak ?

stoty:
I have pushed this by accident, but then force-pushed the previous state.
So everything should be OK, you don't have to do anything wrt the bad push I made @zabetak .

zabetak:
@stoty What command are you using to check if OWASP plugin runs fine or not?
```
./gradlew dependencyCheckAnalyze -PnvdApiKey=xxxxxx
./gradlew dependencyCheckAggregate -PnvdApiKey=xxxxxx
```
I have tried the commands above and both fail with NPE on every JDK that I tested.

stoty:
I use

`./gradlew -PnvdApiKey=REDACTED dependencyCheckAggregate` 

@zabetak 

I've just tested it with this patch on JDK11 successfully.

> openjdk version "11.0.28" 2025-07-15
> OpenJDK Runtime Environment (build 11.0.28+6-post-Ubuntu-1ubuntu124.04.1)
> OpenJDK 64-Bit Server VM (build 11.0.28+6-post-Ubuntu-1ubuntu124.04.1, mixed mode, sharing)

zabetak:
It also worked yesterday but for the past few hours it consistently fails or hands; quite strange.

zabetak:
Today the errors disappeared also for me. I run `./gradlew dependencyCheckAggregate -PnvdApiKey=xxx` with `/opt/jdks/zulu21.42.19-ca-jdk21.0.7-linux_x64` and it passed (although it took a while).

zabetak:
Rebased, squashed, and amended commit message. I will merge this once CI runs.

PR #296, opened 2025-09-11
[CALCITE-7177] Update Guava from 33.4.0-jre to 33.4.8-jre in Avatica

richardantal:


richardantal:
@F21 and @zabetak What do you think about Istvan's suggestion about shading or excluding these?
Should we create a follow up ticket for this?

zabetak:
At the moment we are not relocating Guava so it doesn't make sense to relocate `org.jspecify.annotations`, that is a transitive dependency. The `org.jspecify.annotations` does not seem to be an optional dependency so I guess its not safe to remove/exclude it either. Due to this I am +1 with the current state of the PR.

From my perspective any kind of shading or relocation discussion can be deferred to a follow-up.

F21 (approved):

PR #310, opened 2026-06-02
avoid running static initializers in CursorFactory.fromProto

sahvx655-wq:
CursorFactory.fromProto resolves a wire-supplied class name with Class.forName(String), which initializes the class and runs its static initializer. The server hits this while parsing an ExecuteRequest from an untrusted client (StatementHandle -> Signature -> CursorFactory), so a client can make the server load and initialize any class on its classpath. Resolve with the three-arg form and initialize=false; the class is still linked but no static code runs at parse time.

### DIFF
--- a/core/src/main/java/org/apache/calcite/avatica/Meta.java
+++ b/core/src/main/java/org/apache/calcite/avatica/Meta.java
@@ -722,7 +722,10 @@ public interface Meta {
 
       if (proto.hasField(CLASS_NAME_DESCRIPTOR)) {
         try {
-          clz = Class.forName(proto.getClassName());
+          // Resolve without initializing: a server parsing this from an untrusted
+          // client must not run the named class's static initializer.
+          clz = Class.forName(proto.getClassName(), false,
+              CursorFactory.class.getClassLoader());
         } catch (ClassNotFoundException e) {
           throw new RuntimeException(e);
         }


F21:
Can you please log the issue in JIRA @ https://issues.apache.org ?

sahvx655-wq:
Sure, I'll raise it on the Apache JIRA and put the CALCITE key in the PR title so it's tracked properly. For the writeup I'll describe what I found while tracing the server parse path: CursorFactory.fromProto resolves the wire-supplied className with the single-arg Class.forName, which loads and initialises the class in one go. That path is reachable on the server straight from an untrusted ExecuteRequest, through StatementHandle.fromProto then Signature.fromProto then CursorFactory.fromProto, so a client can name any class on the server's classpath and have its static initialiser run at request-parse time.

Left unfixed, the concern is initialisation side effects firing before the class is ever legitimately used, which is the usual CWE-470 unsafe class loading case. The change keeps the three-arg Class.forName with initialize=false so the class is still linked and resolved, and genuine initialisation still happens lazily when a cursor actually gets built. I'll post the JIRA link back here once it's raised.

mihaibudiu:
This PR has a conflict, can you please fix it?

mihaibudiu:
Apparently this is a duplicate of [CALCITE-7610] so maybe we can close it?

mihaibudiu:
@zabetak did the work on 7610

zabetak:
This is a duplicate and has a more complete fix in CALCITE-7610 thus closing the PR.

PR #314, opened 2026-08-04
[CALCITE-7494] Avatica conversion to string of TIMESTAMP WITH TIME ZONE does not include time zone

mihaibudiu:


mihaibudiu:
This affects the Calcite output, causing CI to fail.

mihaibudiu:
I guess I should first submit a helper PR in Calcite to disable the failing tests, and after merging this we can re-enable them.

mihaibudiu:
Whoever is reviewing this, please also check the failing CI test output, since it shows what the Calcite results will be in the future if we merge this.

xiedeyantu:
As I understand it, the issue is simply that Avatica was losing the "UTC" designation, so the goal of this PR (including the tests) is to verify that the fix preserves that designation. Therefore, there is no need to verify the correctness of the time zone handling itself (such as in the example below)—is that correct? 
```
select TIMESTAMP WITH TIME ZONE '2020-01-01 00:00:00 America/New_York';
+---------------------+
| EXPR$0              |
+---------------------+
| 2020-01-01 05:00:00 |
+---------------------+
```

mihaibudiu:
Yes, this is right. This will need to be merged in 3 steps: disable calcite tests, merge this, enable calcite tests

xiedeyantu (approved):

F21:
@mihaibudiu I can see that @xiedeyantu has approved this PR. Can you go ahead with the required merges to get this in?

mihaibudiu:
@F21 once we merge https://github.com/apache/calcite/pull/5247 we can hopefully merge this as well.

PR #316, opened 2026-08-11
[CALCITE-7708] Avatica does not support MAP types

mihaibudiu:


mihaibudiu:
Draft until [CALCITE-7710] is solved.

PR #318, opened 2026-08-13
CALCITE-7715 Remove HttpCore Timeout from BuiltInConnectionProperty

exceptionfactory:
CALCITE-7715 Removes the HttpCore 5 `Timeout` class reference from the `BuiltInConnectionProperty` classes and replaces it with the equivalent number of milliseconds to set the default Connection and Response Timeout properties.

This change keeps references to HttpCore and HttpClient 5 localized to the Avatica Remote Client classes. With this change, integrating projects that do not need to perform remote HTTP operations with Calcite Avatica can exclude the HttpComponent dependencies without any issues at runtime.


mihaibudiu:
Can't you also remove the unused dependency?

mihaibudiu:
BTW: In our project we use Calcite but not Avatica connections, and excluding the http components explicitly in the Java build worked just fine, even without this change.

exceptionfactory:
> Can't you also remove the unused dependency?

Thanks for asking. The dependency itself is still required, so an intermediate step could be to mark the HTTP Components dependencies as optional, but that would require consuming projects to explicitly include them, instead of explicitly exclude them.

mihaibudiu:
In this case I am not sure that this change matches the justification in JIRA. As I said, if you don't use http, you can actually remove the dependencies already.

exceptionfactory:
> In this case I am not sure that this change matches the justification in JIRA. As I said, if you don't use http, you can actually remove the dependencies already.

The change is required for excluding those dependencies because `BuiltInConnectionProperty` is referenced when loading standard `Driver` and `UnregisteredDriver` classes, so the reference to `Timeout` causes class reference issues at runtime. Removing this reference enables the runtime exclusion strategy, and should also provide the foundation for future potential improvements along these lines.

mihaibudiu (approved):

exceptionfactory:
Thanks for the review and feedback @mihaibudiu!

PR #320, opened 2026-08-26
[CALCITE-7745] Implement a mechanism to restrict the connection properties that can be set by the caller

rubenada:
[CALCITE-7745] Implement a mechanism to restrict the connection properties that can be set by the caller
Implemented an allowlist / denylist mechanism (disabled by default, i.e. everything is allowed, for backwards compatibility)

### DIFF
--- a/server/src/test/java/org/apache/calcite/avatica/jdbc/JdbcMetaTest.java
+++ b/server/src/test/java/org/apache/calcite/avatica/jdbc/JdbcMetaTest.java
@@ -32,14 +32,19 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.fail;
 
 /**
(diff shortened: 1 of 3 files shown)


server/src/test/java/org/apache/calcite/avatica/jdbc/JdbcMetaTest.java
```diff
@@ -202,6 +207,122 @@ public class JdbcMetaTest {
...
+    assertThat(e.getMessage(),
+        containsString(JdbcMeta.CLIENT_PROPERTIES_ALLOWLIST_KEY));
+  }
+
+  @Test public void testCheckClientPropertiesAllowlistPermitsListedName() {
+    JdbcMeta.checkClientProperties(null, setOf("propA", "propB"),
+        Collections.singletonMap("propA", "v"));
+  }
+
+  @Test public void testCheckClientPropertiesDenylistAppliedBeforeAllowlist() {
```
thomasrebele: The name of the test is a bit misleading: if there are several properties, there may be an allowlist exception for one property, even if there is a denylist exception for another property. This depends on the iteration order.

How about `testCheckClientPropertiesDenylistAppliedBeforeAllowlistForCertainProperty`?

server/src/test/java/org/apache/calcite/avatica/jdbc/JdbcMetaTest.java
```diff
@@ -202,6 +207,122 @@ public class JdbcMetaTest {
...
+  }
+
+  @Test public void testCheckClientPropertiesNullMapBypassesChecks() {
+    JdbcMeta.checkClientProperties(setOf("propA"), setOf("propA"), null);
+  }
+
+  @Test public void testParsePropertyNameListNullOrBlankIsUnconfigured() {
+    // A null or all-blank value means the rule was not configured; the
+    // parser must return null so the check treats it as absent, not as an
+    // empty allowlist that would reject everything.
```
thomasrebele, line 283: If someone wants to reject all properties, how would they do it? Or is there a property that the client needs to always set?

server/src/test/java/org/apache/calcite/avatica/jdbc/JdbcMetaTest.java
rubenada: Ok, changed.

server/src/test/java/org/apache/calcite/avatica/jdbc/JdbcMetaTest.java
rubenada: Typically some properties should always be needed (like user, password, schema, etc).
The mechanism is based on backwards compatibility, so everything keeps working as before (i.e. everything allowed by default). I guess if someone would really want to reject all properties, it could be theoretically done by defining one, single, "dummy" property on the allowlist, so de facto all real properties will be rejected.

thomasrebele: Ok, we can keep it as it is. The "dummy" could be put on both, allowlist and denylist, so that no properties are accepted.

thomasrebele (approved):
LGTM

rubenada:
Thanks for the review @thomasrebele !

